Issue #284, opened 2017-05-16
On admin page when user is select looks like you are logged out


cjmyers:
When looking at the user admin page, it looks like you are logged out.

16zzundel5:
PR #285 ready for review.

PR #285, opened 2017-05-17
Fix error with logged-in view on users page

16zzundel5:
Issue was that I changed the name user to currentUser, changed that back in users.js and renamed the colliding variable name in users.jade.


16zzundel5:
This is addressing the bug described in #284

Issue #220, opened 2017-03-24
Make sort and search for Collection view work


cjmyers:

udp:
Hidden the options for now pending implementation serverside.

16zzundel5:
PR #323

cjmyers:
This appears to be working well.  Let’s leave this issue open though to record the fact that we might want to still add “full search” support.

cjmyers:
Closing.

PR #323, opened 2017-05-25
Add searching and sorting to collection table view

16zzundel5:
Addresses #220


cjmyers:
This appears to be working well.  Let’s leave this issue open though to record the fact that we might want to still add “full search” support.

PR #447, opened 2017-08-24
Updated Footer with Github Link. Also http -> https

ethanwillis:


cjmyers:
wiki.synbiohub.org is not on https, but on http.  This should be changed back to http.

zachzundel:
Changed!

ethanwillis:
HTTPS would be better if possible. I could help with setting up letsencrypt on your mediawiki install.

zachzundel:
The server is managed by Newcastle University, so the decision is theirs.  (@udp) We used Let's Encrypt for the reference instance (synbiohub.org)

cjmyers:
I believe we decided there is nothing on the wiki that needs encrypting.  Our synbiohub instance which actually communicates part information is using https, and that is the more critical thing for security.

udp:
I'm quite happy to enable HTTPS for the wiki.  The http link would redirect anyway, so it wouldn't affect this issue.  (Also, we should probably move the wiki out of Newcastle at some point.)

cjmyers:
As long as http still works too.  Recall we are using wiki URLs for some of our annotations, and they all use http.  Would be a pain to change now.

PR #514, opened 2017-11-09
ZIP Upload feature

zachzundel:
This pull request adds two main features:

1. Improved submission page UI
2. Support for ZIP uploads


zachzundel:
I think that the collection text is good to have outside the info button, but it was a bit wordy to show every time. I edited it down and moved it back to the submit page. It's one extra line and offers a bit of help to those who may be daunted even by our small info buttons.

PR #526, opened 2017-11-22
Don't assume accept header is present

schnauzer:
Some tools (golang http stdlib) don't send 'Accept: */*', so
default to the api endpoint if it's not there.


zachzundel:
Thanks!

cjmyers:
@zachzundel do we need to make this change other places?

zachzundel:
I don’t think so, we usually test for whether or not something explicitly disallows html and default to giving the HTML version.

PR #532, opened 2017-11-27
A new look for the view pages

cjmyers:
1) Buttons are shrunk to be just icons with tooltip text to provide more information.  When there are multiple items with the same icon, it is converted into a menu (such as for Download on the Component page).  This allows the buttons to be put onto a row instead of a column eliminating the problems with text running into buttons.  This also allows the Collection icon to be moved to the right of the page avoiding collection names crashing into icons.

2) Organized the sections of the page into collapsable panels.  This prevents the more detailed sections from cluttering the page especially for those who are not so interested in all the various annotations.  By default the “Details” panel is open initially.

### DIFF
--- a/public/styles/site/synbiohub.less
+++ b/public/styles/site/synbiohub.less
@@ -85,7 +85,7 @@ a {
 }
 
 .entry-actions .btn {
-    width: 200px;
+    width: 40px;
     text-align: left;
 }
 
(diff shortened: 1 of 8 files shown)


udp:
You appear to have duplicated the work I did in the new design branch?  I agree with these changes, but it’s important we work together...

cjmyers:
Not exactly.  I did look at that one for inspiration, and I took a less radical redesign approach.  I did not change the overall layout at all, but I’m open to idea of moving top options into side menu as you suggest.  However, I think you need to add return to Home and Search to the side menu in that case.

PR #559, opened 2018-01-10
Update any `file:` links we can

zachzundel:


zachzundel:
Yes I tested it. It is a DELETE INSERT WHERE query that is executed together. 

Thanks,
Zach

cjmyers:
Ok, nice.  Will try it out.

cjmyers:
Appears to be working.  I’ve merged it.  Let’s get Meher to test it further.

PR #576, opened 2018-02-06
Autodocker

zachzundel:
This PR enables automatic snapshot Docker images. Every commit triggers a Docker build which is then pushed to the `synbiohub/synbiohub:latest` tag on Docker Hub. It does not yet support autotagging of new releases, but that isn't a high priority at the moment since we just did a release and have some time before the next one.


zachzundel:
Sorry, changed it so that the snapshot branch tag is `synbiohub/synbiohub:snapshot` instead of `synbiohub/synbiohub:latest`, thought that was clearer.

PR #583, opened 2018-02-21
User Profile Page

zachzundel:
Adds basic user profile page which allows them to change their full name, email, university affiliation, and password.

### DIFF
--- a/templates/views/profile.jade
+++ b/templates/views/profile.jade
@@ -2,8 +2,50 @@
 extends ../layouts/default.jade
 
 block content
-    div.container-fluid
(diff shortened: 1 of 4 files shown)


templates/views/profile.jade
```diff
@@ -2,8 +2,50 @@
 extends ../layouts/default.jade
 
 block content
-    div.container-fluid
-        div.row
-            div.col-md-12
-                h1 #{user.username}
```
a4ff7810: Maybe it'd still be useful to include the username somewhere else on the page?

PR #591, opened 2018-03-01
Feature/sbol attachments

zachzundel:
Adds support for upload and download of SBOL attachments, and migrates SBH attachments to SBOL attachments.


lib/views/attachment.js
```diff
@@ -120,72 +120,73 @@ module.exports = function(req, res) {
...
-	    
         locals.meta = meta
 
-	locals.rdfType = {
-	    name : 'Attachment',
-	    url : 'http://wiki.synbiohub.org/wiki/Terms/SynBioHub#Attachment'
-	}
+        locals.rdfType = {
+            name: 'Attachment',
+            url: 'http://wiki.synbiohub.org/wiki/Terms/SynBioHub#Attachment'
```
cjmyers: The URL should change to SBOL URL

cjmyers: Hmm, actually I think our intention is to put these definitions on our wiki.

zachzundel:
Are we ready to merge this into master? I've rebased it so the merge should be (relatively) painless.

cjmyers:
Probably close.  Let me test a bit more.

Issue #646, opened 2018-07-06
Add a ability during setup to set the URI prefix


cjmyers:
Add to the Setup screen the ability to set the URI prefix.  There can be a big warning that says this is usually the same as the URL unless you are using this instance for debug purposes to pretend to be a different instance.  

The URI prefix should be used to set the following configurations in config.local.json:

  "databasePrefix": "\<URI prefix\>",
  "triplestore": {
 ...
    "defaultGraph": "\<URI prefix\>public",
    "graphPrefix": "\<URI prefix\>",
 ...
  },

Note that the URI prefix should end with a "/", or one should be added.  Finally, the URI prefix should be used to create the graph URI for the admin user that is created during setup, i.e.,

\<URI prefix\>user/\<adminId\>/

zachzundel:
Ready for review #672

cjmyers:
Works.

Issue #578, opened 2018-02-08
Unable to change graphPrefix after setup


cjmyers:
There is no way to change the graphprefix after you setup.  When you install a local test instance, one may want the graphprefix and databaseprefix which for some reason are different variables to be something other than localhost:7777.   You can change the databasePrefix but not the graphPrefix.  As far as I can tell these should always be the same, and I’m not sure if the graphPrefix is even used or not.  We should probably sort this one out.

zachzundel:
`grep`ing through the source code shows that there is nowhere that actually uses graphPrefix, so I think I can purge it entirely unless you have objections?

`grep` results:
```
config.json:      "graphPrefix": "http://synbiohub.org/",
config.local.json:    "graphPrefix": "http://localhost:8888/",
lib/views/setup.js:        graphPrefix: settings.instanceUrl,
templates/views/admin/status.jade:                    code #{config.triplestore.graphPrefix}
```

cjmyers:
This issue is not so much about this variable but about the fact that if you forget to set the graph prefix for the public graph correctly in setup, I'm not sure if you can change it later.  However, perhaps you can because if you change the database prefix, it may change the public graph?  Is this correct?

zachzundel:
I'm not sure, I think that if you changed the database prefix, the graphs
would just break.

Zach

cjmyers:
I was not considering moving triples, just retargeting.  I would expect any content in the old graph to no longer be reachable.

Chris

cjmyers:
If you change the database prefix, you must also change the defaultGraph and graphPrefix or you will not be able to access your public graph correctly.

Also, you must change the URI prefix in the SQL database for the users, so they will be using the correct database prefix.  Otherwise, you will have a mix of users using old and new prefix.

cjmyers:
If we only allow this during setup, then it needs to be disabled in admin.

cjmyers:
I think the instanceURL should also be disallowed to change, right?  Also, we don't need the note about ending in a slash, if these cannot be change.  I think the instanceURL and the URI prefix should likely both be displayed on the status page now, since they are not editable.

@michael13162 Not sure while enabling SBOLExplorer is still on the General page.  Shouldn't this be on the SBOLExplorer admin page now?  

@zachzundel I think we should move Allow Public Account Creation checkbox to the Users pages.  This means the General page is now empty and can be removed.

michael13162:
[...]

PR #672, opened 2018-08-24
Spoofing setup

zachzundel:
Make the changes so that spoofing can be done on setup, and not broken after setup.

Addresses issues #578 and #646 .


cjmyers:
Actually, defaultGraph needs to be set to uriPrefix plus "public"

cjmyers:
Committed fix to master

Issue #659, opened 2018-07-30
FROM line in SPARQL interface should be restricted to public and logged in user graphs


cjmyers:
The use of the FROM line in SPARQL needs to only be allowed to specify the public and logged in user graphs.

zachzundel:
Ready for review #684

PR #684, opened 2018-09-04
Check FROM and FROM NAMED on SPARQL queries

zachzundel:
Check FROM requests to make sure that they are only against the user's private graph or the public graph.


zachzundel:
Addresses issue #659

zachzundel:
Tested in meeting with @cjmyers

PR #691, opened 2018-09-05
Issue688

### LINKED ISSUE #688: Paging not working with find uses
When you do find uses, you cannot page and get the right results.

cjmyers:
Fixes #688


cjmyers:
Nope.  Guess I forgot to commit.  Proved your point about reviews :-)

Issue #681, opened 2018-09-01
Setup procedure for upgrading SD2 SBH instance


cjmyers:
1) Setup an automatic update on the staging server to the master branch.

2) Work with the team to develop a testing plan to validate staging before deployment to the main instance.

3) Perhaps update the main instance off of a particular branch.

@zachzundel can you coordinate with Matt Vaughn and the data representation group on this task?

PR #694, opened 2018-09-06
Show revision

zachzundel:
Shows the revision in the footer. Nothing shown if getting the revision fails.


zachzundel:
Cleanup task for #681

PR #698, opened 2018-09-07
Remote ribbon

cjmyers:
This makes remotes have a purple ribbon.  

Note that module definition support in visbol is added, but commented out.  This was a mistake in that I accidentally made my changes while in the branch working on that.


lib/search.js
```diff
@@ -63,7 +63,7 @@ function search(graphUri, criteria, offset, limit, user, explorer = false) {
                     result.url = result.uri.toString()
 		}
 
-                result.triplestore = result.uri.toString().startsWith(config.get('databasePrefix')+'public/')?'public':'private'
+                result.triplestore = result.uri.toString().startsWith(config.get('databasePrefix')+'public/')?'public':result.uri.toString().startsWith(config.get('databasePrefix')+'user/')?'private':'remote'
```
zachzundel: This works, but the ternary operators and the `?` and `:` in the URL make it hard to read. It could probably benefit from being an if/else clause.

cjmyers:
Fair enough.  Updated.

PR #743, opened 2018-12-05
Change persistence to not nest directories

zachzundel:
This fixes the logo reupload problem


zachzundel:
I think there is a subtle difference. The structure create here was:
```
uploads
  '- uploads <- this is the persisted folder, but nothing got put here
``` 
 Whereas I think the top level is persisted in SBOLExplorer, but it creates an extra nested folder. I can look again later.

PR #745, opened 2018-12-06
add testing containers

oflatt:
First try with Travis test.sh script.


oflatt:
Tests are running indefinitely- forgot to wait for a period of time before exiting.

PR #746, opened 2018-12-06
Add missing dependencies for submit

zachzundel:


cjmyers:
Is raptor2 provide raptor ?

zachzundel:
Yep!

PR #750, opened 2018-12-12
Test script that creates a new user

oflatt:
This is the first commit for testing- it simply starts a new instance of SynBioHub and creates a new user.

### DIFF
--- a/tests/createuser.sh
+++ b/tests/createuser.sh
@@ -1,0 +1,34 @@
+source ./testutil.sh
+
+message "Creating initial account"
+
+RESULT=300
+
+while [[ "$RESULT" -ne 302 ]]
+do
+    RESULT=$(curl -X POST \
+                --write-out "%{http_code}" \
+                --silent \
+                --output /dev/null \
+                -F "userName=testuser" \
+                -F "userFullName=Test User" \
+                -F "userEmail=test@user.synbiohub" \
+                -F "userPassword=test" \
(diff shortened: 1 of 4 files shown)


oflatt:
Another question I have is should the cleanup script be run inside of the test script each time it is run? On one hand, the test.sh script won't work correctly without a cleanup first. On the other, I don't want accidental removal of data.

oflatt:
Looks like it failed to find scripts that it depends on. Do I need to get Travis to recognize them as scripts it needs? Or is the path just wrong?

cjmyers:
Should it be tests/createuser.sh?

cjmyers:
I think in test.sh:

# create a new user
 bash ./createuser.sh

should be:

# create a new user
 bash tests/createuser.sh

maybe, not sure, but it seems it cannot find it.  

Good point about local runs.  Maybe we should have an option on the script on whether you want to cleanup on exit.  However, that being said, if you run the test locally without cleanup, then you might fool yourself it is working when it is not.  Best would be to not have the test image and any local image not conflict with each other.  We should likely discuss how to do this when we meet next.  For now, cleanup on exit is probably fine since for now only travis will run this.

zachzundel:
docker-compose has a -p option which lets you specify the project name,
which is what is prepended to the named volumes and container names. You
could specify a test prefix that will let you isolate all the test env from
development

Zach

PR #751, opened 2018-12-16
Test script 2- with correct path and containers in separate project

oflatt:
The script that starts up synbiohub with docker-compose and creates a new user, this time creating the containers under a separate project name.


oflatt:
It seems to be a bash error- I'll try again with #!/bin/bash at the top. Sorry about that.

PR #752, opened 2018-12-16
Start up test script that creates a new user- second try with bash

oflatt:
A test script that starts synbiohub using docker-compose up and creates a test user in the new instance.
It seems like the previous try failed because it was not using bash.


cjmyers:
This error seems the more serious one:

synbiohub_1      | Unhandled rejection SequelizeDatabaseError: SQLITE_ERROR: no such table: user
synbiohub_1      |     at Query.formatError (/synbiohub/node_modules/sequelize/lib/dialects/sqlite/query.js:348:14)
synbiohub_1      |     at afterExecute (/synbiohub/node_modules/sequelize/lib/dialects/sqlite/query.js:112:29)
synbiohub_1      |     at replacement (/synbiohub/node_modules/sqlite3/lib/trace.js:19:31)
synbiohub_1      |     at Statement.errBack (/synbiohub/node_modules/sqlite3/lib/sqlite3.js:16:21)

It appears to be unable to find the user data.

tests/test.sh
```diff
@@ -1,3 +1,61 @@
...
+    git clone https://github.com/synbiohub/synbiohub-docker;
+fi
+
+message "Building SBOLTestRunner"
+cd SBOLTestRunner
+mvn package
+cd ..
+
+message "Starting SynBioHub from Containers"
+docker-compose -f ./synbiohub-docker/docker-compose.yml -p testproject up &
```
zachzundel: Instead of running this in the background with `&`, run with the `-d` option, which puts the containers in the background after they're built.

tests/createuser.sh
```diff
@@ -0,0 +1,34 @@
...
+                -F "instanceURL=http://localhost:7777/" \
+                -F "uriPrefix=http://localhost:7777/" \
+                -F "color=#D25627" \
+                -F "frontPageText=text" \
+                -F "virtuosoINI=/etc/virtuoso-opensource-7/virtuoso.ini" \
+                -F "virtuosoDB=/var/lib/virtuoso-opensource-7/db" \
+                -F "allowPublicSignup=true" \
+                http://localhost:7777/setup)
+    if [[ "$RESULT" -ne 302 ]]
+    then
```
zachzundel: this shouldn't run until you know that SBH is up, so you don't need to cycle. You should fail if you don't get a 302

tests/test.sh
```diff
@@ -1,3 +1,61 @@
...
+while [[ "$(docker logs --tail 1 testproject_synbiohub_1)" != "Resuming 0 job(s)" ]]
```
zachzundel: There are now healthchecks which provide a cleaner way to check whether synbiohub is up. You can get the current health with `docker inspect testproject_synbiohub_1 | jq .[0].State.Health.Status`

tests/test.sh
```diff
@@ -1,3 +1,61 @@
...
+do
+    sleep 5
+done
+
+message "Started successfully"
+
+# wait for logs of other containers
+sleep 30s
```
zachzundel: If SynBioHub is healthy, the other containers should be too. The only exception *might* be ElasticSearch, but you can safely ignore that for now. If we want to test SBOLExplorer we will have to create a healthcheck for ElasticSearch

zachzundel:
Investigating the weird user issue now.

tests/test.sh
oflatt: Thank you for all the help.

tests/createuser.sh
oflatt: Now adding a message "failed to create first account" and exiting with error code one. Is there a way to be more specific?

zachzundel:
Looks good! @oflatt you can merge when you like.

PR #757, opened 2018-12-22
View refactor

cjmyers:
Substantial refactor of the views.  There are now only two view for SBOL:

collection.js
topLevelView.js

The customization is performed using the summarize<Type>.js methods in views/utils directory.  These were built off of sbolmeta, which has now been folded into SynBioHub.

This should not be merged in for a bit still as it needs some substantial testing.


lib/actions/createBenchlingSequence.js
```diff
@@ -2,7 +2,7 @@ var pug = require('pug')
 
 const { fetchSBOLObjectRecursive } = require('../fetch/fetch-sbol-object-recursive')
 
-var sbolmeta = require('sbolmeta')
+var sbolmeta = require('../views/utils/sbolmeta')
```
zachzundel: the refactored sbolmeta should just be a directory in lib

lib/bioschemas/DataRecord.js
```diff
@@ -16,7 +16,7 @@ module.exports = function generateDataRecord(metadata) {
     "mainEntity": {
       "@type": metadata.rdfType,
       "identifier": metadata.name,
-      "hasBioChemRole": metadata.roles.map((role) => role.uri),
+      "hasBioChemRole": metadata.roles?metadata.roles.map((role) => role.uri):[],
```
zachzundel: This operation is too complex for an in-line ternary operation, it is hard to read

lib/views/admin/theme.js
```diff
@@ -70,8 +70,6 @@ function post(req, res) {
 
     localConfig.themeParameters[themeName] = newParameters 
 
-    config.set(localConfig)
```
zachzundel: Why was this deleted?

lib/views/collection.js
```diff
@@ -227,10 +185,16 @@ module.exports = function (req, res) {
             id: metaData.displayId,
             pid: metaData.persistentIdentity,
             wasDerivedFrom: metaData.wasDerivedFrom || '',
-            wasDerivedFroms: metaData.wasDerivedFrom?[ metaData.wasDerivedFrom ]:[],
+            wasDerivedFroms: metaData.wasDerivedFrom?
```
zachzundel: Too complicated for a ternary.

lib/views/collection.js
```diff
@@ -227,10 +185,16 @@ module.exports = function (req, res) {
...
+		[ { uri: metaData.wasDerivedFrom,
+		    url: metaData.wasDerivedFrom } ]:[],
             wasGeneratedBy: { uri: metaData.wasGeneratedBy || '',
 			      url: uriToUrl(metaData.wasGeneratedBy,req)
 			    },
+            wasGeneratedBys: metaData.wasGeneratedBy?
```
zachzundel: Ternaries are hard to read.

lib/views/topLevelView.js
```diff
@@ -0,0 +1,351 @@
+
+const { fetchSBOLObjectRecursive } = require('../fetch/fetch-sbol-object-recursive')
+const { getContainingCollections } = require('../query/local/collection')
+
```
zachzundel: No need to put a space between imports

lib/views/topLevelView.js
```diff
@@ -0,0 +1,351 @@
...
+		})
+            })
+
+        ])
+
+    }).then(function lookupEncodedProteins() {
+
+        var query =
+            'PREFIX sybio: <http://w3id.org/sybio/ont#>\n' +
+        'SELECT ?subject WHERE {' +
```
zachzundel: fix indentation

lib/views/topLevelView.js
```diff
@@ -0,0 +1,351 @@
...
+        return Promise.all([
+	    getContainingCollections(uri, graphUri, req.url).then((_collections) => {
+
+		collections = _collections
+
+		collections.forEach((collection) => {
+
+		    collection.url = uriToUrl(collection.uri)
+
+                    const collectionIcons = config.get('collectionIcons')
```
zachzundel: Could move outside this function so that we do not read from disk/context switch for every collection

lib/views/topLevelView.js
```diff
@@ -0,0 +1,351 @@
...
+                    
+                    if(collectionIcons[collection.uri])
+			collectionIcon = collectionIcons[collection.uri]
+		})
+            }),
+
+	    sparql.queryJson(getCitationsQuery, graphUri).then((results) => {
+
+                citations = results
```
zachzundel: indentation

lib/views/topLevelView.js
```diff
@@ -0,0 +1,351 @@
...
+            encodedProteins = results.map((result) => {
+                return result.subject
+            })
+
+        })
+
+    }).then(function getOtherComponentMetaData() {
+
+        if(meta.protein && meta.protein.encodedBy)
+            otherComponents = otherComponents.concat(meta.protein.encodedBy)
```
zachzundel: if you still have todo here then wrap the if contents in {}

lib/views/topLevelView.js
```diff
@@ -0,0 +1,351 @@
...
+		    }
+
+                })
+
+	    }
+        }
+
+    }).then(function fetchFromIgem() {
+
+	if(topLevel.wasDerivedFrom.toString().indexOf('http://parts.igem.org/') === 0) {
```
zachzundel: Is this hitting the iGem registry on _every_ iGem part pageload? That's a lot, and will be slow. Could we maybe offload this onto a nightly job that updates from iGem?

zachzundel: Or this should be a plugin.

lib/views/admin/theme.js
cjmyers: Not sure why.  Put it back.

lib/bioschemas/DataRecord.js
cjmyers: I think I prefer to keep this one.

lib/views/collection.js
cjmyers: Agreed

lib/views/collection.js
cjmyers: Agreed

lib/views/topLevelView.js
cjmyers: Fixed

lib/views/topLevelView.js
cjmyers: Seems right?

lib/views/topLevelView.js
cjmyers: Would like to move these things out of this file and wrap with only if CD.  Not sure how.

lib/views/topLevelView.js
cjmyers: Would be good to start by moving this out of the file into its own that is only hit on CD

zachzundel:
Tried on some large files from the SBOL test suite, appears to be working

lib/attachments.js
```diff
@@ -98,20 +99,33 @@ function getAttachmentsFromTopLevel(sbol, topLevel, share) {
...
-                type: attachment.format.toString().replace('http://identifiers.org/combine.specifications/','').replace('http://identifiers.org/edam/',''),
-                url: url,
-                size: attachment.size,
-                sizeString: attachment.size == -1 ? null : filesize(attachment.size)
-            })
-        }
+        if(attachment) {
+	    if (!(attachment instanceof URI)) {
+		var url = '/' + attachment.uri.toString().replace(config.get('databasePrefix'), '')
+		if (attachment.uri.toString().startsWith(config.get('databasePrefix') + 'user/') && share) {
```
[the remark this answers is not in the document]
cjmyers: Perhaps.  The tricky thing is how to forward share links.  Namely, if you get a share link to a CD, we also give you share links to all its children and referenced Sequences.

lib/api/datatables.js
```diff
@@ -73,28 +73,10 @@ function collectionMembersDatatable(req, res) {
...
-                    memberType = 'Sequence'
-                } else if (memberType === 'Model') {
-                    memberTypeUrl = 'http://wiki.synbiohub.org/wiki/Terms/SynBioHub#Model'
-                    memberType = 'Model'
-                } else if (memberType === 'Attachment') {
-                    memberTypeUrl = 'http://wiki.synbiohub.org/wiki/Terms/SynBioHub#Attachment'
-                    memberType = 'Attachment'
-                }
+		if (memberType === 'ComponentDefinition') memberType = 'Component'
+		else if (memberType === 'ModuleDefinition') memberType = 'Module'
```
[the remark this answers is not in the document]
cjmyers: Yes, since these are all now handled uniformly.  Namely, the member type is stripped off the end of the type URI, and the URL is at http://sbolstandard.org now, since that is where Goksel's ontology is.

PR #758, opened 2018-12-23
Update readme instructions for manual installation on Ubuntu

oflatt:


oflatt:
It seems I've mixed up the branches accidentally- this one is only supposed to have readme changes.

PR #763, opened 2019-01-02
Linting

zachzundel:
This is a large PR but do not have fear. This adds some linting and style checks. I used the styleguide [here](https://github.com/standard/standard) and ESLint. It also sets up automatic linting and fixing for most issues when you commit, so it should not be a great imposition to development. 

The automatic changes made in this pull request were limited to the subset that would not change the function of the code, so it is safe to merge it into master. The other changes can be made more gradually.

I would also like to set up linting as a PR check step. 

We can wait for #757 and apply it to that code too.


zachzundel:
Oops -- that's only in the browser directory, which it should have ignored. Will fix, one second.

browser/autocomplete.js
```diff
@@ -1,21 +1,21 @@
 
 $('.sbh-autocomplete').typeahead({
-    hint: false,
-    highlight: true,
-    minLength: 1
+hint: false,
```
cjmyers: Why remove the indentation here?

browser/dataIntegration.js
```diff
@@ -7,189 +7,189 @@ var submitEvent
...
+mode: 'list'
 
 }
 
 if(document.getElementById('sbh-data-integration')) {
 
-    h = require('virtual-dom/virtual-hyperscript')
-    clickEvent = require('value-event/click')
-    submitEvent = require('value-event/submit')
+h = require('virtual-dom/virtual-hyperscript')
```
cjmyers: And here?

zachzundel:
Fixed the browser issues.

zachzundel:
This is not urgent, just wanted to start the ball rolling. I agree the view
refactor should get in first.

PR #765, opened 2019-01-04
Update readme for installing on Ubuntu

oflatt:
Major changes include getting rid of the old comsode stuff (which I am told is not necessary anymore) and instructing developers to install Virtuoso from source.


oflatt:
Ah, apparently the readme was updated in the other pull request. Sorry.

PR #768, opened 2019-01-07
Small bugs

cjmyers:


lib/api/datatables.js
```diff
@@ -73,7 +73,7 @@ function collectionMembersDatatable(req, res) {
                     }
                 }
 
-                var memberTypeUrl = member.type.replace('http://sbols.org','http://sbolstandard.org')
+                var memberTypeUrl = member.type?member.type.replace('http://sbols.org','http://sbolstandard.org'):''
```
zachzundel: Shouldn't be a ternary. Or, at least, should have spaces so it's readable.

cjmyers: I'm not as against ternary as you are, but I'm fine making it more readable.  I'll leave that though to the linting later, since I need to get this fix in ASAP.

PR #778, opened 2019-01-14
Python testing infrastructure

oflatt:
I created the methods compare_get_request and compare_post_request that will be used in the future for all endpoint tests.
They check against previously saved results, throwing an error and a diff if the response differs from what is saved. They also throw an error if the saved results do not exist.
Todo: Add support for updating tests for only some endpoints.
Currently, the test suite checks the setup page using a get request and fills in the form using a post request.


tests/test.sh
```diff
@@ -49,8 +57,22 @@ done
...
+fi
+
+message "Running test suite."
+
+# run the set up script
+
+python3 test_suite.py "$@"
+
+
+message "Running SBOLTestRunner"
```
zachzundel: Probably should comment this out until you are running the testrunner

tests/test_functions.py
```diff
@@ -0,0 +1,119 @@
+import requests, difflib, sys
+import re
+
+command_line_arguments = []
+
+number_of_errors = 0
+
+
+def set_command_line_arguments(arguments):
```
zachzundel: look into argparse for command line arg handling

tests/test_functions.py
```diff
@@ -0,0 +1,119 @@
...
+    if "-resetalltests" in command_line_arguments:
+        test_print("resetting all tests by saving every request for future comparisons. You should not have run this unless you are working on the test suite implementation itself and have verified that all tests passed before the reset.")
+
+def test_print(string):
+    sys.stdout.write("[synbiohub test] ")
+    sys.stdout.write(string)
+    sys.stdout.write("\n")
+
+# make html a little more human readable
+def format_html(htmlstring):
```
zachzundel: Try BeautifulSoup for managing HTML (Formatting, parsing, etc.) https://www.crummy.com/software/BeautifulSoup/

tests/test_functions.py
```diff
@@ -0,0 +1,119 @@
...
+    # for an ending angle bracket and then something else, enter a new line
+    newstring = re.sub('>(.)', r'>\n\1', newstring)
+    # for something else then an angle bracket, also insert a new line
+    newstring = re.sub('(.)<', r'\1\n<', newstring)
+    
+    return newstring
+
+# perform a get request
+def get_request(request):
+    response = requests.get('http://localhost:7777/' + request)
```
zachzundel: Not required right now, but tests should allow URL to be specified as a command-line arg

tests/test_functions.py
```diff
@@ -0,0 +1,119 @@
...
+    response.raise_for_status()
+    
+    content = format_html(response.text)
+    return content
+
+# perform and save a get request in previous results
+def save_get_request(request):
+    content = get_request(request)
+
+    filepath = 'previousresults/getrequest_' + request + '.txt'
```
zachzundel: These should be saved as HTML files

tests/test_functions.py
```diff
@@ -0,0 +1,119 @@
+import requests, difflib, sys
+import re
+
+command_line_arguments = []
```
zachzundel: I would just add defaulted parameters for functions where necessary instead of declaring a global variable. 

def function(arg, arg, reset=False)

tests/test_functions.py
```diff
@@ -0,0 +1,119 @@
...
+
+number_of_errors = 0
```
zachzundel: Again this needn't be a global variable, functions should return errors or raise exceptions when they fail and those can be caught by the test

tests/test_functions.py
```diff
@@ -0,0 +1,119 @@
...
+# creates a file path for a given request and request type
+def request_file_path(request, requesttype):
+    return 'previousresults/' + requesttype.replace(" ", "") + "_" + request + ".txt"
+        
+# check a request against previous results
+# request should the page that was requested, such as /setup
+# requesttype is the type of request performed, either "get request" or "post request"
+def compare_request(requestcontent, request, requesttype):
+    # if the global state is to replace all files, do that
+    if "-resetalltests" in command_line_arguments:
```
zachzundel: a function called `compare_request` shouldn't also be writing the request to file

tests/test_suite.py
```diff
@@ -0,0 +1,9 @@
+from sys import argv
+from test_functions import exit_tests, set_command_line_arguments
+
+set_command_line_arguments(argv[1:])
+
+# run the first time setup script
+import firsttimesetup
```
zachzundel: using `import` to run code seems strange

zachzundel:
@oflatt -- I really like where you're going with this, you could probably ignore most of my comments if you like, but some things you're doing could lead to trouble in the future (global variables in Python can do weird things with caching in the Python interpreter in my experience). Good work overall!

tests/test_functions.py
oflatt: Yep- I wrote a todo in the comment at the top of the pull request. It's on my list.

tests/test_functions.py
oflatt: I created a global variable so that it can simply print the error, and then it raises an exception if any of them have failed, with the number of errors it got.
Is there a better way to do this?

tests/test_functions.py
zachzundel: Oh, I meant something different. I mean I should be able to run these tests against `https://synbiohub.org/` or `http://localhost:7371` or any other URL.

tests/test_suite.py
oflatt: I thought that too- but it seems like the best way in Python.

tests/test_functions.py
oflatt: Good idea- I was actually looking at that and planning to do it in the future.
Travis would have to install the python package every time though if I understand correctly.

tests/test_suite.py
zachzundel: I thought you were planning on using something like pytest or other testing framework to organize these? That might make more sense

tests/test_functions.py
zachzundel: I guess what I'm saying is that I think these functions should probably not hold any sort of state between executions. 

You could have code in your `firsttimesetup` module that tracks whether these raise an exception or not. These functions could pass exceptions up to that module. Then you could have a command-line option that specifies whether or the test suite should keep-going in an error case.

tests/test_functions.py
oflatt: Good idea- realized I didn't need the global variable for command line arguments.

tests/test_functions.py
zachzundel: Yep, that's fine. 
Right now, there is an `install` part of the testing stage which just contains a Docker build step. It could include a Python install step very easily.

You would just have to run `pip freeze > requirements.txt` inside your virtualenv and all the currently-installed packages would be documented. Then Travis just has to run `pip install -r requirements.txt`.

tests/test_functions.py
oflatt: Won't need this global in the future.

tests/test_functions.py
oflatt: Oh I see- good idea.

tests/test_functions.py
oflatt: Yes, I agree that it's generally bad practice.
Hmm, not sure what you mean. If it passes the exception back up, then whoever is writing tests has to keep track of exceptions in the response and add to their counter, which I don't like so much, since I am trying to make it as simple as possible to write tests. Is there a way around this?

tests/test_functions.py
oflatt: For the command line option -resetalltests, all the files are saved instead of compared. Would you like me to rename the funtion?

tests/test_suite.py
oflatt: Yeah, I have been looking into pytest and unittest. Not sure yet- I'll look into it more.

oflatt:
All of the requested changes have been implemented besides one- the compare_request function also saves to files if specified in the command line arguments.
You can specify the server to run on and specify which tests to reset in the process.
There are no more global variables because of the switch to unittest.

tests/test_functions.py
```diff
@@ -0,0 +1,106 @@
...
+
+def compare_request(requestcontent, request, requesttype):
+    """ Checks a request against previous results or saves the result of a request.
+request is the endpoint requested, such as /setup
+requesttype is the type of request performed- either 'get request' or 'post request'"""
+    # if the global state is to replace all files, do that
+    if args.resetalltests:
+        with open(request_file_path(request, requesttype), 'w') as rfile:
+            rfile.write(requestcontent)
+    elif requesttype[0:3] == "get" and request in args.resetgetrequests:
```
zachzundel: You could just use `GET` and `POST`, but these substrings aren't a huge issue.

tests/test_requirements.txt
```diff
@@ -0,0 +1,3 @@
+beautifulsoup4
```
zachzundel: Did you create this by hand? Usually they have version numbers, which could be useful for making sure tests don't break in the future because of a package upgrade

oflatt:
I'm still unable to merge this pull because travis won't run. I assume you will look in to it, @zachzundel.
I'm now ready to merge running sboltestrunner in. Should I create a new pull request?

PR #782, opened 2019-01-22
Rendering plugins

zachzundel:
Adds the ability to create and use rendering plugins.

Uses the [igem-plugin](github.com/synbiohub/igem-plugin) three times on [SBH Utah Dev](https://tang.ece.utah.edu/public/igem/BBa_K925000/1) to fetch information from iGEM repository

### DIFF
--- a/browser/plugin.js
+++ b/browser/plugin.js
@@ -1,0 +1,28 @@
+function fetchPluginStream(streamId, $element) {
+    // Make the request, retrying if necessary
+    $.ajax({
+        type: 'GET',
+        url: '/stream/' + streamId,
+        success: function(data) {
+            $($element.closest(".col-md-12")).html(data.body)
+        },
+        statusCode: {
+            503: function(jqXHR) {
+                var retry = jqXHR.getResponseHeader("Retry-After")
+                retry = parseInt(retry, 10) * 1000
+                
+                setTimeout(fetchPluginStream, retry, streamId, $element)
+            },
+            404: function(jqXHR) {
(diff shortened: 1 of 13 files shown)


zachzundel:
There are three different plugins. The configuration of the plugin lets you
do a bit of customization on how the plugin works.

The drop down disappearing is already part of this.

The iGEM plugins has to fetch the SBOL non-recursive to see if it was
derived from an iGEM part.

cjmyers:
Ok, can you add a bit more to the README then for the plugin to describe how to set that up.

I still saw the dropdown with the message “Not Found” inside.

zachzundel:
Sorry -- I've added in the removal of boxes, I forgot to push that. I've also updated the iGEM plugin readme.

cjmyers:
So, I've approved this, but there are still a couple minor things:
1) It would be good to extend the readme on the iGEM plugin to describe completely how to setup the three plugins for iGEM including the full information for config.  I'm also unclear as to whether you need to clone it three times or is there some way to start three instances from one source.  Maybe we could have the start command take an argument as to which config to start with, so we could then checkin the three config files and just start 3 times.
2) The iGEM plugin box does disappear for non-iGEM parts, but it still shows up for the Collection, since that has a wasDerivedFrom that hits an empty iGEM page.  We might want to change that wasDerivedFrom to either point to a page you can render or special case the plugin to skip the Collection page.

Both of these comments though are for the iGEM plugin and not the plugin architecture, which is why I'm approving this.

PR #795, opened 2019-01-26
Hot fix to ignore version number in requests

oflatt:
Now if only one line was changed in a request and that line is the version number, it does not count as a failure.

### DIFF
--- a/tests/first_time_setup.py
+++ b/tests/first_time_setup.py
@@ -13,6 +13,7 @@ class TestSetup(TestCase):
         # fill in the form and submit with test info
         setup = {
             'userName' : 'testuser',
+            'userFullName' : 'Test User',
             'userEmail': 'test@user.synbiohub',
             'userPassword': 'test',
             'userPasswordConfirm': 'test',
--- a/tests/previousresults/postrequest_setup.html
+++ b/tests/previousresults/postrequest_setup.html
@@ -4,33 +4,8 @@
   <meta charset="utf-8"/>
   <meta content="width=device-width, initial-scale=1.0" name="viewport"/>
   <meta content="IE=edge" http-equiv="X-UA-Compatible"/>
-  <meta content="A parts repository for synthetic biology." name="description"/>
-  <script type="application/ld+json">
-   {
-  "name": "Test Synbiohub",
-  "description": "text",
-  "url": "http://localhost:7777/",
-  "keywords": "synbiohub",
-  "identifier": [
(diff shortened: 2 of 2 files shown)


oflatt:
@zachzundel Travis isn't running again, not sure what you had planned.

zachzundel:
Hopefully it should be fixed now. Travis was having issues partially because of a mis-formatted .travis.yml file I didn't catch in the last testing PR

oflatt:
Sorry about that- I forgot to push the changes before creating the pull request.
It now has the functionality described.

tests/test_functions.py
```diff
@@ -73,21 +73,31 @@ def compare_request(requestcontent, request, requesttype):
...
-            if firstchangep:
-                firstchangep = False
+            numofchanges += 1
 
             # print the diff
             changelist.append(c)
             changelist.append("\n")
 
-        if not firstchangep:
+        # if it was only a one line change, there are 11 lines in the diff
```
zachzundel: Can you add an HTML id to the div containing the whole version string, then use BeautifulSoup to empty out the string? That seems cleaner and will be more similar to the approach you'd use to ignore other parts of the page. Writing a regex for each thing to ignore will get unwieldy.

oflatt: That seems like a much better solution- I'll work on that.

oflatt: It seems that an id in html is unique- I will use a class called testignore instead because then multiple items can be of the class. Is that a good way to do it?

zachzundel:
That sounds fine.

Le lun. 28 janv. 2019 à 09:51, Oliver Flatt <notifications@github.com> a
écrit :

oflatt:
It now looks for any html element of the class testignore.

oflatt:
@zachzundel It seems like my change that added a div of the class testignore did not apply when travis ran it. Do you know why?

zachzundel:
It fails on my computer when I run this branch. Are you sure it works for you @oflatt ?

oflatt:
Yeah, I'm sure. I re-ran the same build command as travis and then ran the test.sh on this branch.
Now I'm a bit confused- is there some jade compilation step I'm missing for travis?

zachzundel:
No, the jade is compiled on the fly. I will dig into the issue and see what
I can find. :)

Le lun. 28 janv. 2019 à 20:18, Oliver Flatt <notifications@github.com> a
écrit :

Issue #800, opened 2019-01-30
Unable to View Implementation


nroehner:
After uploading the attached file, I am unable to view the Implementation with displayId "batch_1_sample." Instead, I get a page with "TypeError: identified.getAnnotations is not a function."
[funyes_test.zip](https://github.com/SynBioHub/synbiohub/files/2813071/funyes_test.zip)

cjmyers:
Fixed by PR #801

PR #801, opened 2019-01-30
Imp built bug

cjmyers:
Fixed a bug for example from Nic issue #800

### DIFF
--- a/lib/summarize/summarizeImplementation.js
+++ b/lib/summarize/summarizeImplementation.js
@@ -8,8 +8,12 @@ function summarizeImplementation (implementation, req, sbol, remote, graphUri) {
     return uriToMeta(implementation)
   }
 
+  var built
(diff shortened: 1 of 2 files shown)


lib/summarize/summarizeImplementation.js
```diff
@@ -8,8 +8,12 @@ function summarizeImplementation (implementation, req, sbol, remote, graphUri) {
...
   }
 
+  var built
+  if (implementation.built) {
+    built = summarizeIdentified(implementation.built, req)
+  }
   var summary = {
-    built: summarizeIdentified(implementation.built, req)
+    built: built
   }
```
zachzundel: What will show up if there is no `built`? Will it be empty or throw an error?

cjmyers:
Empty.  This is an optional field.

PR #807, opened 2019-02-01
Warnings for unchecked synbiohub endpoints

oflatt:
This pull request adds checking that all endpoints have been checked. For now, it simply prints warnings.
There are two things that could be done better, but I don't know another way to do them.
1. Global variables are used to keep track of which endpoints have been checked. These variables live in test_functions. However, I'm not sure if this can be avoided without making the code keep track from the outside, but the whole point was to avoid doing that.
2. Parsing app.js to get the lists of endpoints was done using some messy regex. I've looked into some libraries that might help, but this is the solution for now.


oflatt:
this was merged in by #832

PR #818, opened 2019-02-06
Clean up test code to load individual modules in order.

oflatt:
Now modules can depend on the order in which they were run, but test functions within modules should be independent.


oflatt:
Also- this pull depends on #807.

cjmyers:
Looks good to me.  @zachzundel ?

oflatt:
this was merged in by #832

PR #824, opened 2019-02-09
Strange != to !== bug.  The != was true due to type mismatch

cjmyers:
This was a really strange bug.  Linting made us change != to !==, but this caused a bug to appear, since the != did the implicit cast to string allowing a isReplacedByUri.toString() to be actually the thing compared.  Changing to !== now is true because the types are not equal.  This problem may cause other issues wherever we have change != to !==.  Need to be very careful with that.


cjmyers:
@oflatt travis is failing on this pull request.  Not sure why on this one.

PR #826, opened 2019-02-09
Fix fetching for local objects

### LINKED ISSUE #815: Non-local fetch should not fetch locally or for child objects
A non-local fetch should never be performed on the same repository as the original fetch.  Furthermore, fetching child objects is never useful to fetch non-locally, so these should certainly be skipped.

See #808 where this issue was encountered.

cjmyers:
Closes #815


zachzundel:
Okay, I think this is the last of the unsnarling. This one was the messiest so please review the changes and make sure everything is as expected.

cjmyers:
All looks good.  Thanks

PR #827, opened 2019-02-10
Reorganize header layout

### LINKED ISSUE #806: Top right links do not look good with large SBH names and/or small screens
We should perhaps reduce these to icons, so they scale better, or change to some menu that can open.

cjmyers:
Reorganizes header layout to deal with small screens.

Closes #806

### DIFF
--- a/public/styles/site/synbiohub.less
+++ b/public/styles/site/synbiohub.less
@@ -98,6 +98,7 @@ a {
 }
 
 .top-search-form .submit {
+    background-color: #2C3E50;
 	margin-left: 0px;
 }
 
@@ -212,7 +213,7 @@ a {
 .sbol_breadcrumbs ul {
     list-style: none;
     margin: 0;
-    height: 40px;
+    height: 100%;
     -webkit-padding-start: 0px;
 }
 
@@ -385,6 +386,7 @@ a {
 .sbol-breadcrumbs ul li {
     padding: 17px;
     float: left;
+    height: 100%;
 }
 
 .sbol-breadcrumbs ul li a {
--- a/public/styles/site/synbiohub.less
+++ b/public/styles/site/synbiohub.less
@@ -78,12 +78,7 @@ a {
 .login .username {
 }
 
-.top-search-form {
-    margin-left: 50px;
-    margin-top: 1.2em;
-}
-
-.top-search-form input, .main-search-form input {
+.main-search-form input {
     font-size: .9em;
     border: 1px solid white;
     margin-left: -8px;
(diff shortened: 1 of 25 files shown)


cjmyers:
Not quite ideal yet.  There are three magnifying glasses.  I would like to get rid of the one in the search box, but I don't know how.  Check out the search bar on cnn.com that expands out when clicking on the magnifying glass.  I think this looks pretty cool.  Not sure how to execute it though.

cjmyers:
@oflatt This one changes the top bar quite fundamentally, so it will change the HTML results.  If you let me know how to update these for the test, I can do that for this branch.

oflatt:
@cjmyers , if you look at the README in the tests folder, there is a section called "saving results in order to make tests pass." Ask if anything is unclear.
If you are very sure, you can use the --resetalltests option, but I advise against it. Also, make sure to build the docker image if you haven't.

zachzundel:
Did some more rearranging @cjmyers can you take a look?

cjmyers:
Hmm, cannot see them, I tried checking out issue806 and pulling latest, but I think I'm still snarled:
On branch issue806
Your branch is ahead of 'origin/issue806' by 3 commits.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean

Says I'm ahead by 3 commits, which seems strange.

cjmyers:
I reset, but I still don't see your updates.  Did you push them?

cjmyers:
Thanks.  This is going to be a pretty big one, since it changes the header on every HTML page.

zachzundel:
Yeah.... I would delete your branches locally and re-pull them. The command
is git branch -d <branch name>

zachzundel:
Sorry, I thought I had but they failed. They are pushed now.

cjmyers:
Think it might look better on the right side.  Seems slightly more common to find it there on other cites. 

BTW: noticed this move means search is not on every pages.  We should likely sort that too.  Again, right side would look better I think.

zachzundel:
I think that's fine. How do you feel about removing the URI entirely from
the top bar? It doesn't really mean anything to most users. We could have
the whole thing be search, and have it centered.

Zach

udp:
Massive +1 to this.  The URI is already in the address bar; it doesn't 
need to be repeated in the header.

J

zachzundel:
Moved the search bar again, would appreciate input on this revision. :)

cjmyers:
I think it looks good but a couple minor nits:

1) Can we make the page type on left “inactive” and search bar “active”.  This way your chosen theme color shows up better.
2) Can we add a “<icon?> View” to left of the view pages to be consistent with all the other pages.

Otherwise, I think it looks great!

cjmyers:
Looks pretty good, but there is a small amount of active color showing up below the left inactive part.

screen shot 2019-02-11 at 1 33 15 pm

zachzundel:
Mine looks different... which browser are you using?

[image: image.png]

cjmyers:
It is same for me with both Safari and Chrome.

cjmyers:
Looks good now.  Did you fix it?

zachzundel:
I didn’t do anything.

cjmyers:
Mysterious :-)

cjmyers:
We need to update the tests still, since this will cause changes to HTML.

zachzundel:
I already did. :-)

cjmyers:
Brilliant!  Thanks.

PR #828, opened 2019-02-10
Add more search paging

### LINKED ISSUE #812: Search page should have paging buttons at bottom as well as top
The navigation links on the search page (e.g., "Previous 1 2 3 Next") are only at the top of the page.
They should be at the bottom too, so that if you're browsing down through results, you can go to the next page without having to scroll back to the top.

cjmyers:
Closes #812


cjmyers:
This one will change the search page HTML, but I expect it is also failing due to the reorganization of the top bar that got mixed into this one.

templates/views/search.jade
```diff
@@ -105,6 +105,28 @@ block content
...
                     +search-result-component-definition(submission)
+                    
+        div(align='center').search-result-summary
+            | Showing #{firstResultNum} - #{lastResultNum} of #{numResultsTotal} result(s)
+            - var previous = firstResultNum - (limit+1)
+            br
+            if (previous >= 0)
+                a(href=originalUrl + '/?offset='+previous) Previous
+                = ' '
+            - var start = (firstResultNum - (limit*5+1)) < 0               ? 0                     : (firstResultNum - (limit*5+1)) / limit
```
zachzundel: I think some of this could be moved into JavaScript

cjmyers: Perhaps.  It is pretty tricky, so I would prefer not to move that right now.  You can log an issue to move it later.

zachzundel:
Disregard my comments, it looks like the previous implementation was also in Jade. I don't like it much but it should at least stay consistent.

PR #829, opened 2019-02-10
Redirect to collection after submission

### LINKED ISSUE #814: After submission, redirect to something in the submitted file.
Right now, after you submit you get redirected unceremoniously back to the root submissions page. I always want to go check on what I've submitted, though, so it'd be a lot better if I was redirected to something in what I'd submitted.

Possibilities:
1. The first TopLevel in the file
2. The first "root" TopLevel in the file, e.g., a collection that contains many things.

The first should be easy; the second would be harder, but given that most SBOL is machine-generated with the root at the top, might be approximated well by the first anyway.

cjmyers:
Redirects to the submitted object after a successful submit.

Closes #814

### DIFF
--- a/lib/views/submit.js
+++ b/lib/views/submit.js
@@ -5,21 +5,9 @@ var loadTemplate = require('../loadTemplate')
 
 var pug = require('pug')
 
-var retrieveCitations = require('../citations')
-
 var fs = require('mz/fs')
 
-var async = require('async')
-
-var SBOLDocument = require('sboljs')
-
 var extend = require('xtend')
-
-var uuid = require('uuid')
-
-var serializeSBOL = require('../serializeSBOL')
-
-var request = require('request')
 
 var config = require('../config')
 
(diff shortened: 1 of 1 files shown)


lib/views/submit.js
```diff
@@ -509,7 +497,7 @@ function submitPost (req, res) {
             if (req.forceNoHTML || !req.accepts('text/html')) {
               res.status(200).type('text/plain').send('Successfully uploaded')
             } else {
-              res.redirect('/manage')
+              res.redirect('/user/' + encodeURIComponent(submissionData.createdBy.username) + '/' + collectionId + '/' + collectionId + '_collection/' + version)
```
zachzundel: I can't remember -- can you submit into a public collection? If you can, would this break?

cjmyers: Submission must go to a private collection.

zachzundel:
Moved the search box into the left side of the screen. @cjmyers what do you think of this?

PR #833, opened 2019-02-11
Revert "Redirect to collection after submission"

cjmyers:
Reverts SynBioHub/synbiohub#829


zachzundel:
I don't think this is necessary any more.

PR #834, opened 2019-02-11
Fix overwrite_merge for fields

cjmyers:
This fixes the issue with overwrite_merge being changed for fields by mistake.


zachzundel:
Added a change that fixes submit entirely without needing to revert the previous PR. Please check it out @cjmyers

PR #836, opened 2019-02-12
Append to log file

zachzundel:
Modify the `tee` command so that it always appends to the log


cjmyers:
Approved but lets merge in after 1.4.1 release as part of 1.4.2 along with a better logging procedure.

PR #844, opened 2019-02-21
More robust plugin failure handling

### LINKED ISSUE #825: Plugins that are down are not handled gracefully
When a plugin cannot be reached, it should gracefully timeout on trying to reach the plugin.

zachzundel:
If a plugin is down, an error message is logged and the box is deleted from the page. 

Also includes lint fixes for `lib/views/topLevelView.js`.

Closes #825

### DIFF
--- a/lib/api/stream.js
+++ b/lib/api/stream.js
@@ -2,44 +2,42 @@ const uuid = require('uuid/v4')
 
 let responses = {}
 
-function serve(req, res) {
-    let id = req.params.id
+function serve (req, res) {
+  let id = req.params.id
 
-    switch (responses[id]) {
-        case undefined:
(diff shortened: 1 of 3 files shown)


cjmyers:
Will this continue to hit the plugin in subsequent page loads?  Also, this appears to still leave the iGEM code for now, and we will replace with plugin with a later PR, correct?

zachzundel:
Yes. Every page load will trigger a plugin request but as soon as the connection cannot go through the page will stop trying to pull the data. This is better than permanently disabling because it handles transient failures. 

Yes, a future PR will remove all the iGEM code. Replacing with a plugin will require setup on an individual SBH instance, won't be part of a PR.

PR #853, opened 2019-02-26
Remove user parameter from getGraphUriFromTopLevelUri

zachzundel:
A first step in pulling auth into its own layer. Part of the process of reducing surface area of user information.


zachzundel:
This is one step in making the authentication and authorization system more modular. There are lots of places that use information about the currently authenticated user where another method could be used to the same effect. If I can reduce the 'awareness' of each page of the ins and outs of the auth system, then it can be more easily refactored into its own layer. One of the common uses of the request user was in the getGraphUriFromTopLevelUri function, so I read and re-wrote the function to do the same thing without examining the currently-authenticated user.

zachzundel:
Good catch, I didn't realize this was all that was blocking someone from accessing another graph. Closing this PR and will submit as part of the larger changes.

PR #854, opened 2019-02-27
Fix removal of collection owner

zachzundel:
While I was refactoring out authentication I found some issues with the remove owner functionality


zachzundel:
Didn't mean to leave them in but will keep them there as I am also working on a PR that converts all the logging statements to using log-levels as we have discussed offline.

cjmyers:
Ok

PR #860, opened 2019-03-07
Move to leveled logs

zachzundel:
- Sets up Winston as a log management system
 - Separates logs into graded files in a logs/ directory
 - Cycles the logs daily
 - Replaces the console.\* functions with their winston equivalent

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@ synbiohub.log
 public/local/*
 unzipped/*
 jsbml.log
-
+logs
 
 *~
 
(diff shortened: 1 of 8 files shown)


cjmyers:
This looks good, but a few questions:
1) What is shown on the download log page now?
2) Do we still need to change all the console.log calls?  If so, to what?
3) Has this been tested?

zachzundel:
It shows all log messages from “info” and more serious that that.

console.log, console.warn, and console.err calls are already swapped out
for their logger equivalents (this happens in synbiohub.js).

I have tested it locally, I’m not sure what tests it needs in Travis.

PR #861, opened 2019-03-08
Move to button for updating explorer index

oflatt:
New endpoint for updating the index as well.


oflatt:
Also, please take a look at app.js. I made some changes to make the linter happy, I think they are okay but not quite sure on the !== one.

Issue #864, opened 2019-03-13
When converting GenBank/FASTA files may need to uniqueify identifiers


cjmyers:
Some software like SnapGene give the same LOCUS identifier (Exported) to all exported GenBank files.  This is problematic, since this is used as the SBOL identifier, and if multiple files are uploaded they will all have the same identifier.  In this case, should uniqueify the identifiers.

cjmyers:
Perhaps based on the file names.

zachzundel:
Can this be closed @cjmyers ?

cjmyers:
Yes, this was fixed.

Issue #862, opened 2019-03-09
JBEI-ICE ACS and PUBLIC removed 


jurquiza:
Just wondering if Synbiohub public will have ICE-ACS and ICE-PUBLIC collections back?

cjmyers:
It looks like their SSL certificates are expired.  I've let them know.

cjmyers:
This is working again on the development instance (https://dev.synbiohub.org <https://dev.synbiohub.org/>).  Please check it out.

It will be working again on main instance after next update.

cjmyers:
This is fixed

PR #866, opened 2019-03-15
Use filename for GenBank/FASTA as default displayId

cjmyers:
Fixes issue #864

### DIFF
--- a/browser/synbiohub.js
+++ b/browser/synbiohub.js
@@ -132,8 +132,8 @@ if (typeof meta !== 'undefined') {
         processing: true,
         serverSide: true,
 
-        searching: !meta.remote,
-        ordering: !meta.remote,
+      searching: true, //!meta.remote,
+      ordering: true, // !meta.remote,
 	order: [[2, "asc"]],
         
 	ajax: {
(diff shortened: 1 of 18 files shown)


cjmyers:
Also fixes #862

browser/synbiohub.js
```diff
@@ -141,7 +141,7 @@ if (typeof meta !== 'undefined') {
             type: 'GET',
             data: function (d) {
                 d.type = 'collectionMembers'
-                d.collectionUri = meta.uri
+                d.collectionUri = meta.remote?meta.uri.toString().replace("/1","/current"):meta.uri
```
zachzundel: Can you reformat the ternary so that there are spaces between the operands? Did this not raise a linting error?

nodemon.json
```diff
@@ -4,7 +4,8 @@
         "config.json",
         "config.local.json",
         "browser",
-        "public/styles/theme"
+        "public/styles/theme",
+        "node_modules"
```
zachzundel: I don't think it is wise to watch node_modules. Creating watches on each file is expensive and can make development difficult on some machines.

cjmyers: Agreed, these are directories/files to ignore watching.  The watch on node_modules was causing be problems.

zachzundel: Oh, sorry, my bad.

browser/synbiohub.js
cjmyers: Done.  No linting did not complain about it.

zachzundel: Oh, right, it doesn't link the browser directory.

cjmyers:
@zachzundel BTW: I will be checking into this branch until Travis is fixed.  Sorry that it clutters it, but the travis broken is a big issue and they have not yet replied to my query.

zachzundel:
Okay. Why not check into different branches?

Zach

cjmyers:
Because I need all the changes to come together and I cannot get them from master into my other branches.

PR #873, opened 2019-03-31
Test that data persists after restarting docker containers

oflatt:
The tests create a user in addition to the admin, and uploads three collections. The tests that check the views on these collections are run before and after restarting containers.


oflatt:
Looks like it's failing due to another date. My last approach to this kind of problem was to add a tag to html that should be ignored, so I'll look into doing something similar.

oflatt:
Looks like not the travis build is also getting the "forEach of undefined" error that I was seeing-
are the docker containers that travis pulls out of date?

cjmyers:
Hmm, should not be.  They get built automatically with pulls to master.

cjmyers:
I wonder if your branch is out-of-date.  Could you try rebasing this branch to master?  You might also see what happens locally when you checkout this branch.

oflatt:
I'll try the rebase. Locally it works correctly.

oflatt:
I ran yarn upgrade and uploaded my local yarn.lock. This seems like the right thing to do unless it updated a dependency it wasn't supposed to.

oflatt:
It seems like it is just the ordering of data in meta.types- not sure how to fix this one yet

oflatt:
Is it possible that the sortByNames function in the lib/views/advancedSearch.js file is going wrong?

PR #876, opened 2019-04-01
Add support for Measure class

cjmyers:


lib/summarize/summarizeMeasure.js
```diff
@@ -0,0 +1,42 @@
...
+    var sboPrefix = 'http://identifiers.org/biomodels.sbo/'
+    if (type.toString().indexOf(sboPrefix) === 0) {
+      var sboTerm = type.toString().slice(sboPrefix.length).split('_').join(':')
+      typeResult = { uri: type.toString(),
+        term: type.toString(),
+        description: systemsBiologyOntology[sboTerm]
+      }
+    } else {
+      typeResult = { uri: type.toString()
+      }
```
zachzundel: Did this pass lint checks? Looks strange

cjmyers:
It is an abstract type. Not sure. I put identified and top level there. Maybe view is ok

Chris

cjmyers:
Ah, I did not put in layout.  I put componentInstance in views, so I put measured in views.  It is a bit inconsistent.  Do you think we should put all abstract classes in layout?

PR #879, opened 2019-04-01
Fix loading the config early on first start in docker

oflatt:
The change is to do that same change as #767 to the recursive version of the module.
More ambitiously, merging the two modules or making a submodule they depend on would be good, since both seem very similar. It might prevent errors like this.


oflatt:
Restarting build due to the transient error we have been observing.

PR #885, opened 2019-04-09
Add new administrator sparql interface

cjmyers:
New SPARQL update panel on admin page.


lib/app.js
```diff
@@ -283,6 +285,22 @@ function App () {
   app.post('/admin/federate', requireAdmin, bodyParser.urlencoded({ extended: true }), actions.admin.federate)
   app.post('/admin/retrieveFromWebOfRegistries', requireAdmin, actions.admin.retrieve)
 
+  app.get('/admin/sparql', requireAdmin, sparqlAdmin)
+  app.post('/admin/sparql', requireAdmin, bodyParser.urlencoded({ extended: true }), sparqlAdmin)
+
+  function sparqlAdmin (req, res) {
```
zachzundel: Can this get moved down near the sparql function it is copied from?

cjmyers: Sure.  I actually moved it up to be with the other admin pages though.  Which do you think is better?

lib/views/admin/sparql.js
```diff
@@ -47,3 +75,57 @@ module.exports = function (req, res) {
...
+  } catch (e) {
+    form(req, res, {
+      query: req.body.query,
+      graph: req.body.graph,
+      errors: [
+        e.stack
+      ]
+    })
+
+    return
```
zachzundel: This isn't necessary

lib/views/admin/sparql.js
```diff
@@ -47,3 +75,57 @@ module.exports = function (req, res) {
     })
   }
 }
+
+function post (req, res) {
+  req.setTimeout(0) // no timeout
+
+  var graphUri
```
zachzundel: If we declare and them immediately set (like we do here) this can be combined.

lib/app.js
zachzundel: Wherever they are (I am partial to below all the routes, but this is a preference) they should be together, I think.

cjmyers: Ah, you mean just the function?  Yes, I can move that.

zachzundel: Yep, just the `sparqlAdmin` function.

cjmyers:
Made the suggested changes.

PR #889, opened 2019-04-16
Update synbiohub test suite dependency on jq

oflatt:


oflatt:
The fail looks like the same ordering bug that I thought was fixed. Investigating.

oflatt:
Still a bit lost. Maybe meta.types comes from somewhere else, not advancedSearch?

PR #896, opened 2019-04-21
Updated to SBH 1.5, update libSBOLj 2.4.0, commented out search for ice

cjmyers:


cjmyers:
This is an update of version number to 1.5, and update libSBOLj to 2.4.0.  There is some commented out code that I checked in for my start on ICE search.

PR #901, opened 2019-04-24
Remove sliver dependency

zachzundel:
Removed dependency that was causing issues on tang.


cjmyers:
I approve this but it looks like it is running into a similar sort issue with testing.  This time on line 34 of docker_test.py.  You will need to comment out this test likely to get it through.  I guess we are going to need to sort this sort issue soon.

PR #902, opened 2019-04-25
Remove iGEM code replicated in plugin

zachzundel:
Remove the code that fetches from iGEM because the plugins are ready to go


cjmyers:
We should probably move the footer to 1.5.1, before we deploy any of these new changes.

PR #904, opened 2019-04-25
Print error log when tests fail

oflatt:
Currently, prints 50 lines of the error log each time a test fails.

### DIFF
--- a/tests/fetch_logs.py
+++ b/tests/fetch_logs.py
@@ -15,14 +15,40 @@ def file_tail(filename, length):
             return "".join(lines)
 
 
+def log_to_year_month_day(log):
+    list1 = log.split("-")
+    list1[-1] = list1[-1].split(".")[0]
+    return (int(list1[1]), int(list1[2]), int(list1[3]))
+
+
+# returns true if log1 has a newer date than log2
+def newer_logp(log1, log2):
+    date1 = log_to_year_month_day(log1)
+    date2 = log_to_year_month_day(log2)
+    for i in range(len(date1)):
+        if date1[i] > date2[i]:
+            return True
+        elif date1[i] < date2[i]:
+            return False
+    # they must have been the same date to reach this point
+    return False
+
 
 # if num_of_lines is negative, returns all lines
 def get_end_of_error_log(num_of_lines):
-    copy_docker_log()
+#    copy_docker_log()
     directory = os.listdir("./logs_from_test_suite")
+    possible_error_logs = []
     for filename in directory:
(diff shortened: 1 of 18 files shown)


tests/test_functions.py
```diff
@@ -223,3 +226,31 @@ def cleanup_check():
     Checks to make sure all endpoints were tested."""
 
     test_state.cleanup_check()
+
+
+def run_bash(command):
+    process = subprocess.Popen(command.split(), stdout=subprocess.PIPE)
+    output, error = process.communicate()
+
+def file_tail(filename, length):
```
zachzundel: Why not read the whole file into memory and then print the lines? failure in os.popen can be hard to handle

oflatt: Good idea- I was probably over-optimizing

tests/run_sboltestrunner.sh
```diff
@@ -38,5 +38,7 @@ java -jar SBOLTestRunner/target/SBOLTestRunner-0.0.1-SNAPSHOT-withDependencies.j
 
 exitcode=$?
 if [ $exitcode -ne 0 ]; then
```
zachzundel: Does SBOLTestRunner exit with an error code if it fails I’m not sure it does.

tests/fetch_logs.py
```diff
@@ -0,0 +1,38 @@
...
+    process = subprocess.Popen(command.split(), stdout=subprocess.PIPE)
+    output, error = process.communicate()
+
+
+# if length is negative, returns all lines
+def file_tail(filename, length):
+    with open(filename, "r") as file_to_read:
+        lines = file_to_read.readlines()
+        if length < len(lines) and length >= 0:
+            return "".join(lines[len(lines)-length:])
```
zachzundel: Does this print new lines or does it squish everything onto the same line?

tests/run_sboltestrunner.sh
oflatt: I have tested it and looked at the code, I believe it does.

oflatt:
A new problem is that I need to parse the date on the error logs to find the newest one to print.

tests/fetch_logs.py
oflatt: This prints new lines.

zachzundel:
Would it be easier to use `docker logs testsuiteproject_synbiohub_1` ?

oflatt:
Probably, but it's nice to have a local version of the logs. I've implemented the search for the newest log.

PR #905, opened 2019-04-28
Download plugins

zachzundel:
This adds the ability to install plugins which take SBOL and turn it into arbitrary data for user download.


zachzundel:
You can run it with an iGEM plugin if you would like to see it working.

Zach

PR #906, opened 2019-04-29
Sort

zachzundel:
This applies a sort to the result of summarizeSBOL. All objects are sorted according to their JSON representation. This is not the ideal sort from a usability point of view but will provide stability until this can be augmented to provide better sorting.


cjmyers:
Looks like a good temporary solution.

cjmyers:
Let's make sure to restore the tests that were breaking before merging this.

PR #910, opened 2019-05-01
Fix the docker config for backup

zachzundel:
Now the virtuoso directory is mounted in docker-compose so the container can access backups.


cjmyers:
Ok, but we should have an issue to try to figure out how to do backups properly in the docker install.  Even if this is just documenting how it is done.

PR #926, opened 2019-05-04
Remove topLevel and owner tags on submit

### LINKED ISSUE #924: sbh:topLevel Not Removed From Uploaded Objects
After copying an object that I downloaded from SynBioHub into a new namespace, I uploaded the copy to SynBioHub and re-downloaded it, only to discover that the object now had two sbh:topLevel properties, While this could be avoided by changing our library copy methods to remove sbh properties, this seems like an issue that should be fixed on the SynBioHub end by removing existing sbh properties from uploaded objects before creating new ones.

cjmyers:
Fixes #924 

When uploading a file downloaded from SBH, it may have topLevel and ownedBy tags leftover.  So, these are removed on submit now before adding new ones.


cjmyers:
This PR should not be merged until the performance issue is debugged.

PR #933, opened 2019-05-16
Add basic editing for roles

zachzundel:
This is very basic support for editing roles. It does not support adding or removing roles. The URL of the desired role must be pasted into the box.

### DIFF
--- a/lib/api/editObject.js
+++ b/lib/api/editObject.js
@@ -5,7 +5,8 @@ const sparql = require('../sparql/sparql')
 
 const PREDICATES = {
   'title': 'http://purl.org/dc/terms/title',
-  'description': 'http://purl.org/dc/terms/description'
+  'description': 'http://purl.org/dc/terms/description',
+  'role': 'http://sbols.org/v2#role'
 }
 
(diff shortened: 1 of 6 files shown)


lib/api/editObject.js
```diff
@@ -22,6 +23,15 @@ function objectIsValid (object) {
   return object && object !== ''
 }
 
+function tripleIsValid (subject, predicate, object) {
+  if (predicate === PREDICATES['role']) {
```
zachzundel: This requires additional validation. I am not sure how to do this. @cjmyers -- ideas?

cjmyers: The only validation required is that the provided role is a URI.  I wonder if you try to convert the string entered to a URI, if maybe it will give you an exception if the string is not a valid URI string.

zachzundel:
Thanks for the idea @cjmyers -- it took some work to implement but you now see the URI for the ontology term, but it uses the code from summarizeRoles.js (which I pulled out into its own file outside of summarize) to decide what term to display when it finishes.

PR #935, opened 2019-05-18
Add "modified" to updated constructs

### LINKED ISSUE #934: Update Modified Date Attribute when item is updated
When a SynBioHub entry is edited on the web site, the modified date attribute of the the should be updated.

zachzundel:
Closes #934


cjmyers:
Actually, there is a problem with this.  It adds extra modified dates rather than replaces the existing ones.

PR #937, opened 2019-05-21
URL attachments

### LINKED ISSUE #913: Allow URL attachments
Per @jakebeal 's request

3ach:
An initial attempt at URL attachments.

Closes #913

### DIFF
--- a/browser/field-editor.js
+++ b/browser/field-editor.js
@@ -1,11 +1,24 @@
-function updateSparql(value, field) {
+function updateSparql(value, field, cb) {
     console.log(window.location.href + "/edit/" + field)
     $.ajax({
         type: 'POST',
         url: window.location.href + "/edit/" + field,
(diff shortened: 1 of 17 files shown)


cjmyers:
Looks good though having a dropdown for common formats would be useful.  Maybe we could mine the formats that exist in the triplestore and provide those as alternatives to start with.

zachzundel:
Do you mean dynamically offering the option of attaching formats that only exist in that SynBioHub's triplestore, or examining all the triplestores we can to get a list of attachment types?

cjmyers:
I was thinking of something similar to advanced search that populates dropdown with what is in the data.  Using web-of-registries to get more options could be good too.  Key is will still need option to provide a URI when it does not appear.

Alternative is to just give all choices from EDAM/MediaTypes, but that would be a long list.

PR #950, opened 2019-06-13
Should not sort displayList or it reorders glyphs

cjmyers:


cjmyers:
Good point.  Also, probably need a way to test VisBol renderings too.

PR #957, opened 2019-06-28
No rendering

### LINKED ISSUE #771: Report when VISBOL rendering is omitted for being too large
When a VISBOL rendering is omitted for being too large, it should be reported on the page, since otherwise the person would expect it to be there.

Similar warning messages might also want to be done for other places where we don't want silent failure, but obviously not for things that have no inherent rendering.

cjmyers:
Fixes message on large objects to mention no VisBol rendering.  Fixes #771 

Also, updates version number to 1.5.2, and changes a couple log messages.

### DIFF
--- a/lib/views/admin/sparql.js
+++ b/lib/views/admin/sparql.js
@@ -100,6 +100,8 @@ function post (req, res) {
 
   const queryString = generator.stringify(query)
 
+  console.log(queryString)
+
   sparql.updateQueryJson(queryString, graphUri).then((results) => {
     let headers = new Set()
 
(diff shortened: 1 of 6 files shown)


lib/views/admin/sparql.js
```diff
@@ -100,6 +100,8 @@ function post (req, res) {
 
   const queryString = generator.stringify(query)
 
+  console.log(queryString)
```
3ach: Could this be console.debug or other more verbose level?

cjmyers: Done

Issue #948, opened 2019-06-07
MapsTo doesn't appear on page


jakebeal:
MapsTo relations don't appear anywhere on a SynBioHub page.  They should appear under the FunctionalComponent or Module in some way (or if that's too difficult, in their own box). I know #47 says this should be handled by visualization, but right now it is not, and we still need some way of finding the information.

Example:
https://synbiohub.org/user/chanben/Ben_Testing/md0/1.0.0/18cb15adb6919bdbb287633d24ffd2dd54ff3c24/share

cjmyers:
They are children of children, so they appear on the child page, see:

https://synbiohub.org/user/chanben/Ben_Testing/md0/template/1.0.0/b2e51a9cbd8cf7ac8f46dccc9e4c26580bbaaf1e/share <https://synbiohub.org/user/chanben/Ben_Testing/md0/template/1.0.0/b2e51a9cbd8cf7ac8f46dccc9e4c26580bbaaf1e/share>

jakebeal:
I see --- I somehow failed to locate it before. 

Nevertheless, that makes it quite difficult to track down and validate the information, since to actually see the MapsTo you have to go not just into the FunctionalComponent but down another level into the Map, which then gives you Components completely without context. Can't we do the same sort of thing as with Interactions, and show the substructure on the ModuleDefinition page where it actually makes sense?

cjmyers:
Yes, we could do that.  The issue we’ve struggled with is the push/pull of overloading pages and leaving out things.  There could be a lot of these, and it could get really busy.  These are associated with both FunctionalComponents and Modules, so they would need to be listed under these.  Actually, I find the Interactions list a bit busy when it is a long list.

jakebeal:
The principle that I would espouse for deciding what goes on a page is to draw the boundaries based on TopLevel vs. Child: all children of a TopLevel should be on the same page as the TopLevel, since they're defined in order to correctly describe the TopLevel.

cjmyers:
Mostly agree, but we cannot show full information about children on a single page.  However, having some information like Interaction/Participation could be done.

cjmyers:
How's this? @jakebeal 
<img width="1161" alt="Screen Shot 2019-06-28 at 11 23 29 AM" src="https://user-images.githubusercontent.com/3507191/60336047-316f4d00-9997-11e9-91ab-2bd78945508a.png">

jakebeal:
[...]

PR #958, opened 2019-06-28
Add mapsTos to the ModuleDefinition page

cjmyers:
Fixes issue #948

### DIFF
--- a/templates/views/moduleDefinition.jade
+++ b/templates/views/moduleDefinition.jade
@@ -40,21 +40,40 @@ block topLevelPanels
                                 table.table.table-striped.entry-detail-table
                                     tr
                                         td 
-                                            | Instance
+                                            | Instance : 
                                             a(href='http://sbolstandard.org/v2#Module',title='Learn more about Modules') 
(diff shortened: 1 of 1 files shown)


templates/views/moduleDefinition.jade
```diff
@@ -40,21 +40,40 @@ block topLevelPanels
...
-                                        td
-                                            for module in meta.modules
+                                                | &nbsp;
+                                                | :
+                                                | &nbsp;
                                                 a(href=module.definition.url) #{module.definition.name}
+                                                if module.mapsTos.length > 0
+                                                  | (
+                                                  - var first = true
+                                                  for mapsTo in module.mapsTos
```
3ach: This should be in JavaScript, not jade.

cjmyers: I know what you mean, and it was tricky to figure out how to do it.  However, moving to Javascript would mean precomputing the display string.  Not bad I guess.  I could be moved to summarize perhaps.  I think I'm going to merge this for now though.

PR #959, opened 2019-06-28
Add a back to root collection button

cjmyers:


lib/getUrisFromReq.js
```diff
@@ -23,6 +25,7 @@ function getUrisFromReq (req, res) {
       designId = designId + '/' + req.params.version
     }
     url = '/user/' + encodeURIComponent(req.params.userId) + '/' + designId
+    baseUrl = '/user/' + encodeURIComponent(req.params.userId) + '/' + req.params.collectionId
```
3ach: These strings are a bit confusing, they could be rewritten to have subparts clearly shown in larger parts:
```
let baseUrl = '/user/' + encodeURIComponent(req.params.userId) + '/' + collectionId
let url = baseUrl + '/' + req.params.displayId /* maybe + version */
let baseUri = config.get('databasePrefix') + baseUrl
```

etc. etc. Not really a requirement but could make the relationship between these parts easier to understand.

tests/previousresults/getrequest_public-:collectionId-:displayId(*)-:version-full_.html
```diff
@@ -78,7 +78,7 @@ <h5>
         <h5>
          Upload New Icon
         </h5>
-        <form action="/public/testid0/testid0_collection/1/icon" class="form" enctype="multipart/form-data" id="iconForm" method="POST">
+        <form action="undefined/icon" class="form" enctype="multipart/form-data" id="iconForm" method="POST">
```
3ach: Shouldn't be undefined.

tests/previousresults/getrequest_public-:collectionId-:displayId(*)-:version-full_.html
```diff
@@ -320,39 +326,39 @@ <h4 class="panel-title">
            <tbody>
             <tr>
              <td>
-              synbiohub#topLevel
-              <a href="http://wiki.synbiohub.org/wiki/Terms/synbiohub#topLevel" title="Learn more about this annotation type">
+              synbiohub#ownedBy
```
3ach: Is this a test reordering? We need to be careful checking in such changes (also need to fix them.)

cjmyers:
Should be fixed now.  Also was not showing icon on collection page which is also fixed now.

PR #960, opened 2019-07-01
Don't crash on missing file

### LINKED ISSUE #954: Missing attachment file crashes SBH hard
If the attachment file cannot be found when you select to download an attachment, SBH crashes hard and must restart.

I discovered this with an attachment uploaded to DEV then tried to download from SBH.

3ach:
Instead of letting an error event get to the top level and kill node, check if the file exists and throw an exception to be caught when an attachment is missing.

Closes #954

### DIFF
--- a/lib/api/download.js
+++ b/lib/api/download.js
@@ -1,17 +1,10 @@
-var pug = require('pug')
-
 const { fetchSBOLObjectRecursive } = require('../fetch/fetch-sbol-object-recursive')
-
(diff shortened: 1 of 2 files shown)


cjmyers:
This does not look complete.  I only see one file changed and it seems to just be linting fixes.  Did you forget to check-in a file?

cjmyers:
Failing tests.  You may need to update them or check out why.

3ach:
Updated @cjmyers .

PR #961, opened 2019-07-01
Turn off update on public parts

### LINKED ISSUE #942: Pencils still showing up for public parts
Public parts should not allow name/description/roles or other SBOL elements to be edited, so the pencil should not be visible on public parts.

3ach:
Closes #942


cjmyers:
Failing tests, so likely need to check that out.

cjmyers:
Web of registries was down.  Bringing that up and it now passes.

3ach:
Sorry, that might have been my fault. Thanks for fixing it

Issue #786, opened 2019-01-23
Update Advanced Search


jakebeal:
When working with a large unreleased collection on SynBioHub, it would be useful to be able to search that collection using the standard search tools (if it is one that you own, of course).

cjmyers:
That is already supported.  Are you not seeing these results?  If not, can you provide an example search query?  Is this via the browser or API?

jakebeal:
I'm using the browser, and attempting to find media with an advanced search.  I plug in "http://purl.obolibrary.org/obo/NCIT_C85504" into "Role" and it tells me there is nothing that matches that role.

cjmyers:
Ah, I see the issue now.  The Advanced Search UI is actually an area of SBH that really needs some attention.  It was developed for some iGEM demonstrations than abandoned.  You can get what you want via the API.  You can also get it by typing this into the Search box:

role=<http://purl.obolibrary.org/obo/NCIT_C85504>&

This will call this endpoint:

https://hub.sd2e.org/search/role%3D%3Chttp%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCIT_C85504%3E%26

The problem with Advanced Search is that it populates a list of Roles based on those found in the public repository.  If you had even one part in the public repo with this Role, then you would see this option.  

So a simple solution would be to populate this list from both the public graph and your private graph.  This seems reasonable and simple, so we will definitely make this fix.  

One issue though will be that we want these items in the list to be human readable.  For SO terms, we have an OBO file that gives us a nice human readable term.  However, for your role, we would likely end up just calling it NCIT_C85504 rather than Growth Medium, which would be much nicer.  This fact has likely been part of what has stalled advanced search improvements as we weren't sure the best way to generalize this to not be so iGEM specific and make it generic but still human friendly.  Thoughts on this part are very welcome.

jakebeal:
I'd like to suggest starting by using the non-human-readable terms, then allowing the human-readable ones to be added later.  I'd always want to be able to plug in the specific URI if desired, in order to be able to distinguish between different ontologies that use the same human-readable term.

cjmyers:
Seems reasonable.  Maybe we can chat about an Advanced Search redesign next week.  It is really iGEM centric at the moment.  The key question is how to find a good middle ground between basic search and SPARQL.

cjmyers:
This type of query is now supported in the new advanced search.  Closing.

PR #964, opened 2019-07-05
Adv search update

cjmyers:
New generalized advanced search.  Try it out.


cjmyers:
Fixes #786

cjmyers:
Good point.  I’ve switched to using mixin.

PR #975, opened 2019-07-11
Catch a bad submit

cjmyers:


lib/views/submit.js
```diff
@@ -336,7 +336,7 @@ function submitPost (req, res) {
               uri: uri
             }
             removeQuery = loadTemplate('sparql/remove.sparql', templateParams)
-            sparql.deleteStaggered(removeQuery, graphUri).then(doSubmission)
+            return sparql.deleteStaggered(removeQuery, graphUri).then(doSubmission)
```
udp: Couldn't have done it better myself.

cjmyers: Haha, thanks

PR #979, opened 2019-07-21
Log restarts

### LINKED ISSUE #967: Add a message to all logs when the system restarts
It would be useful to see when the system restarts as this can help find the location of fatal bugs.

3ach:
To identify when SynBioHub restarts for debugging

Closes #967


cjmyers:
Hmm, this one is also failing but log looks different.

PR #980, opened 2019-07-21
Add remove attachment button

### LINKED ISSUE #931: Can't delete attachments on page where they have been added.
Once an attachment has been added, it cannot be deleted from the page where it has been added. It took me a while to realize that I had to click through to the attachment, delete it directly (which doesn't make it obvious that it will delete the relationship), then lose my navigation as I get bounced to the root collections page again.

It would be much simpler to have a little "X" to let you delete attachments on the page where you are looking at them.

(discovered while playing with things relating to #913)

3ach:
This button removes attachments on the screen of the object they are attached to.

Also fixed a very serious security bug in that a permission issue would be reported, but the remove would continue anyway.

Closes #931

### DIFF
--- a/browser/synbiohub.js
+++ b/browser/synbiohub.js
@@ -1,9 +1,6 @@
 
 $(document).on('click', '[data-uri]', function () {
-
     window.location = $(this).attr('data-uri')
-
-
 })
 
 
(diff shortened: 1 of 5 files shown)


cjmyers:
That does sound nasty bug.  Changes look fine.

PR #990, opened 2019-07-29
Show "Image" as type, if the attachment is an image format

### LINKED ISSUE #989: Image list not getting populated in wiki editor for mutable fields
This list is populated with attachments of Image type, since the type field had other values than Image in them, then these images were not showing up in the list.   Now, the types are analyzed and replaced with Image, if it is an image type.

cjmyers:
Fixes #989

### DIFF
--- a/lib/summarize/summarizeAttachment.js
+++ b/lib/summarize/summarizeAttachment.js
@@ -1,4 +1,3 @@
-var summarizeTopLevel = require('./summarizeTopLevel')
 
 var config = require('../config')
 var sha1 = require('sha1')
@@ -47,6 +46,8 @@ function summarizeAttachment (attachment, req, sbol, remote, graphUri) {
       attachment.format.toString().indexOf('imageAttachment') >= 0 ||
       attachment.format.toString().indexOf('image') >= 0
   }
+
+  var summarizeTopLevel = require('./summarizeTopLevel')
 
   return Object.assign(summary, summarizeTopLevel(attachment, req, sbol, remote, graphUri))
 }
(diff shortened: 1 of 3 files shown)


lib/summarize/summarizeAttachment.js
```diff
@@ -48,6 +47,8 @@ function summarizeAttachment (attachment, req, sbol, remote, graphUri) {
       attachment.format.toString().indexOf('image') >= 0
   }
 
+  var summarizeTopLevel = require('./summarizeTopLevel')
```
3ach: Why not do the require at the top so it’s clear it’s an external dependency?

templates/mixins/attachments.jade
```diff
@@ -13,7 +13,7 @@ mixin attachments(url, attachments, canEdit)
             tbody
                 for attachment in attachments
                     tr
-                        td #{attachment.type}
+                        td #{attachment.image?'Image':attachment.type}
```
3ach: This could be done in JS and doing so would allow us to create other broad categories.

lib/summarize/summarizeAttachment.js
cjmyers: This was a strange one.  I've noticed with recursion that these require variables can get messed up.  The only solution I have found has been to reset them in the body of the code.

templates/mixins/attachments.jade
cjmyers: Agreed.  It is a bit of clugge.  The format types for attachments need a bit of work, especially when it comes to EDAM types.

PR #992, opened 2019-08-05
fix #986 in config

### LINKED ISSUE #986: Schema.org markup uses wrong namespaces
Markup embedded on [home page](https://synbiohub.org/) uses the bioschemas namespace, e.g. `https://bioschemas.org/DataCatalog`. It should instead use schema.org for those types that are in the schema.org vocabulary, e.g. `https://schema.org/DataCatalog`.

Schema.org types:
- DataCatalog
- Dataset

Bioschema types (being developed for inclusion into schema.org):
- BioChemEntity (and various subtypes for Gene, Protein, etc)
- DataRecord

You'll also need to look at the various properties that Bioschemas are proposing.

udp:


cjmyers:
You will need to update the tests to merge this, you can find instructions here on how to do that:

PR #1001, opened 2019-09-01
Use local fetch for public graph

cjmyers:
I discovered that local public objects were being fetched as if their were non-local.  This greatly impacted fetch performance.

### DIFF
--- a/lib/fetch/local/fetch-sbol-object-recursive.js
+++ b/lib/fetch/local/fetch-sbol-object-recursive.js
@@ -182,7 +182,7 @@ function getSBOLRecursive (sbol, type, graphUri) {
                 prefix = uri.toString()
               }
             }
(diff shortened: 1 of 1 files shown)


lib/fetch/local/fetch-sbol-object-recursive.js
```diff
@@ -349,10 +349,12 @@ return [
 '}'
 ]).join('\n') */
 
+  var from = 'FROM <' + config.get('databasePrefix') + 'public>' + ' FROM <' + graphUri + '>'
+
   var query = [
     isCount
-      ? 'SELECT (count(?s) as ?count) WHERE {'
-      : 'CONSTRUCT { ?s ?p ?o } WHERE {'
+      ? 'SELECT (count(?s) as ?count) ' + from + ' WHERE {'
```
3ach: Do you not have to use a FROM keyword here?

PR #1003, opened 2019-09-14
Timer warning from dependency

3ach:
The timer warning was coming from a dependency, this updates that to get rid of it.


cjmyers:
Noticed that travis died on this one because we exceeded maximum time limit on jobs.  We may need to think a bit about our testing to make sure we don't take too long to run our tests.  Perhaps, we don't need to run the entire SBOLTestSuite.

PR #1004, opened 2019-09-15
Improve the efficiency of Travis

3ach:


3ach:
Yes, it still needs some work, but it does two things. It makes it so
containers are only built once and then shared between the stages, and it
runs the sbol test suite and the SBH test suite in parallel instead of
sequentially.

Zach

cjmyers:
Sounds good.

Chris

PR #1018, opened 2019-10-02
tests endpoint advancedSearch

jvscholz:
tests endpoint advancedSearch

### DIFF
--- a/tests/test_logout.py
+++ b/tests/test_logout.py
@@ -1,0 +1,15 @@
+import requests
+from unittest import TestCase
+from test_functions import compare_get_request, compare_post_request
+
+
+class TestLogout(TestCase):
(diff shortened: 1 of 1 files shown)


cjmyers:
Says you have conflicts with tests/test_logout.py.

Also, why do I see the test but no test results being checked in.

jvscholz:
Sorry, this is just a test pull. I should have commented that. For some reason when I resetgetrequests the reset doesn't get made, but when I run tests/test.sh it passes, so I was seeing if the tests would fail if run on travis

cjmyers:
Ah, I see.

cjmyers:
Is this one obsolete?

jvscholz:
Yes, this one is obsolete

PR #1035, opened 2019-10-22
tests endpoint search/query

jvscholz:
tests endpoint search/query


cjmyers:
There are no search results.  Can you modify the query to return some search results?

jvscholz:
My thinking was that the search results may change depending on updates or what repositories are being run locally so the test would fail- therefore you get around it by searching something with no results. If that logic doesn't hold I'll search something.

cjmyers:
The search results should not change often, since you are only searching locally, and it will only find things you put into the database.

jvscholz:
Updated

cjmyers:
There are still no search results.  A query of "123" will not generate any results.  Try it on the test system and see what query will generate results.

jvscholz:
Is there a way to import the same Design Collections as synbiohub? Otherwise, I don't think the same search results will appear and the test will differ from the previous result, causing the test to fail. Right now there are no search results because no collections exist. Thanks

cjmyers:
The test already does submit and make public at least one Component, so as long as you do a search that finds that one component, you will get consistent search results.  Run the test with --stopaftertestsuite, and you will see there is a public Collection.

jvscholz:
Updated

PR #1043, opened 2019-11-05
tests endpoint autocomplete/query

jvscholz:
tests endpoint autocomplete/query


cjmyers:
Actually, if there was more than one item in your database, it might return something.  Have you tried it on synbiohub.org for example?

jvscholz:
Yes, every part that I've searched using both utah.synbiohub and synbiohub has retured []

PR #1049, opened 2019-11-06
tests endpoint publiccollection_copyfromremote

jvscholz:
tests endpoint publiccollection_copyfromremote


cjmyers:
There is a conflict that I'm not sure how to resolve.

cjmyers:
Seems you might have a copy and paste error

jvscholz:
Updated

cjmyers:
Not really going to do much, since you have no remotes, but okay.

PR #1050, opened 2019-11-06
Added references section

JMante1:


cjmyers:
I'd like us to refine this a bit to be a bit more precise on how to reference this.  We also need to consider that this is going to go on all pages, and not all pages describe parts.  Therefore, it should it either be generalized in the text OR specialized for different page types.

PR #1051, opened 2019-11-07
Tests endpoint public/createSnapshot

jvscholz:


cjmyers:
I'm not sure what this one is supposed to do, and if it really still works.  Maybe hold off on this one till we meet.

PR #1056, opened 2019-11-08
tests endpont public/part/similar

jvscholz:


cjmyers:
As per our discussion, this actually exposes a bug which is similar should return an error code when SBOLExplorer is not running.

PR #1059, opened 2019-11-12
tests endpoint user/part/addOwner

jvscholz:


cjmyers:
Looks like you might have some files mixed from a different tests (sbol/sbolnr).

PR #1060, opened 2019-11-13
tests endpoint user/part/createTest

jvscholz:


cjmyers:
Ok, but this is not perhaps completely functional right now.

PR #1061, opened 2019-11-13
Revert "tests endpoint public/collection/collectionid/displayid/sbol"

jvscholz:
Reverts SynBioHub/synbiohub#1046


cjmyers:
Ok, but we should put this back together with the libSBOLj comparison.

PR #1064, opened 2019-11-25
adds libSBOLj support within testing suite

jvscholz:


cjmyers:
Looks like you are checking in the libSBOLj source code.  Please don't do that.  Your test script should clone and build, but you should put this checked out directory into .gitignore.

PR #1065, opened 2019-11-25
add libsbolj support to testing suite

jvscholz:


cjmyers:
This branch appears to still have the /sbolnr test in it, which fails because SBOL files are not ordered and you need to use libSBOLj to do the comparisons.  Please double check that you remove this test or better yet use the built libSBOLj to do the comparison.

PR #1082, opened 2019-12-19
New status message when use SBOLExplorer is unchecked

yu-eric:


cjmyers:
Seems you might not have checked in the explorerOff.jade

cjmyers:
Will need to update tests too on this one.  James or I can show you how to do this.

PR #1095, opened 2019-12-28
Refactor app.js and support download from URI

cjmyers:
Refactored app.js and removed deprecated endpoints.  This pull request also changes functionality of API for object URI endpoints.  Namely, it returns html when accept header is text/html, and it returns SBOL otherwise.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -200,6 +200,7 @@ function App () {
 
   initSSE(app)
 
+  // Data integration experimental endpoints created by James McLaughlin
   if (config.get('experimental').dataIntegration) {
     app.get('/jobs', requireUser, views.jobs)
     app.post('/actions/job/cancel', requireUser, actions.cancelJob)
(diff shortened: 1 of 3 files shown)


lib/app.js
```diff
@@ -200,13 +200,18 @@ function App () {
...
     app.get('/jobs', requireUser, views.jobs)
     app.post('/actions/job/cancel', requireUser, actions.cancelJob)
     app.post('/actions/job/restart', requireUser, actions.restartJob)
     app.get('/admin/jobs', requireAdmin, views.admin.jobs)
     app.all('/user/:userId/:collectionId/:displayId/:version([^\\.]+)/integrate', requireUser, views.dataIntegration)
     app.all('/public/:collectionId/:displayId/:version([^\\.]+)/integrate', views.dataIntegration)
+    app.get('/user/:userId/:collectionId/:displayId/:version/createImplementation', requireUser, actions.createImplementation)
+    app.get('/user/:userId/:collectionId/:displayId/:version/createTest', requireUser, actions.createTest)
+    // TODO: need to decide if createSnapshot is functional and should be kept or not
+    app.get('/public/:collectionId/:displayId/:version/createSnapshot', actions.createSnapshot)
```
udp: How come these are merged into the dataIntegration section?

lib/app.js
```diff
@@ -200,13 +200,18 @@ function App () {
 
   initSSE(app)
 
+  // Data integration experimental endpoints created by James McLaughlin
```
udp: We can probably ditch these now.

lib/app.js
cjmyers: I put them there as they are experimental and likely in most cases not actively developed.  I think createTest and createImplementation really never got quite finished and don't use the new SBOL classes.  I'm not sure the status of createSnapshot.  Thought it might be best to leave them in until we decide what to do with them, but not make them exposed in the release version.

udp: Ok, fine by me

lib/app.js
cjmyers: We can leave them as experimental, if you think we might want to resurrect this code at some point.  I don't recall exactly what they did.

Issue #628, opened 2018-05-20
Stack trace when trying to copy a collection


palchicz:
Hello, 

I was just playing around with features on Synbiohub and ran into a stack trace.

**Steps to recreate**
1. Navigated to the "manage submissions page" on synbiohub https://synbiohub.org/manage
2. Identified a collection (an export of the collection can be found here:
[Tryingoutthisthing_collection.xml.zip](https://github.com/SynBioHub/synbiohub/files/2019919/Tryingoutthisthing_collection.xml.zip)) I wanted to copy and clicked the copy button.
3. Updated the ID to "try2", the version to "2", and selected the "Prevent submission" option
4. Hit the copy button

**Expected Behavior**
I would get a new collection that was a copy of the old one with an updated Id and version

**Actual Behavior**
No new copy was created and I got the following stack trace:

```
Error: java.lang.NullPointerException
	at org.sbolstandard.core2.SBOLDocument.removeTopLevel(SBOLDocument.java:4399)
	at org.sbolstandard.core2.SBOLDocument.removeCollection(SBOLDocument.java:479)
	at org.synbiohub.CloneSubmissionJob.execute(CloneSubmissionJob.java:83)
	at org.synbiohub.Main$1.run(Main.java:30)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)

    at responseCallbacks.(anonymous function) (/home/deploy/synbiohub/lib/java.js:89:24)
    at DestroyableTransform.javaProcess.stdout.pipe.on (/home/deploy/synbiohub/lib/java.js:51:5)
    at emitOne (events.js:96:13)
    at DestroyableTransform.emit (events.js:188:7)
    at readableAddChunk (/home/deploy/synbiohub/node_modules/readable-stream/lib/_stream_readable.js:211:18)
    at DestroyableTransform.Readable.push (/home/deploy/synbiohub/node_modules/readable-stream/lib/_stream_readable.js:170:10)
    at DestroyableTransform.Transform.push (/home/deploy/synbiohub/node_modules/readable-stream/lib/_stream_transform.js:123:32)
[...]

Issue #261, opened 2017-04-27
Need to double check annotations are updated properly by clone


cjmyers:
When copying, need to ensure isMemberOf and topLevel annotations are being updated correctly.

cjmyers:
Fixed by PR #1101

PR #1101, opened 2020-01-01
Make manage collection endpoints more API friendly

cjmyers:


cjmyers:
Also fixes issue #628 and #261 

Still not entirely sure about cloneSubmission endpoint.  When id is changed, then it is essentially copying everything into a new collection with new URI prefix.  When the version is changed, it only changes the version of the collection.  This behavior is a little strange, but I'm not confident on changing version of the objects as always being safe. 

Tempted to remove the cloneSubmission feature, if there is no compelling use case.  The original use case was to rename things to avoid name clashes before making public.  Still can be an issue, but not sure this sorts that or not.

cjmyers:
Yeah, sort of.  The big issue I have is in what actually should have their URIs updated, and what purpose this is for.  It basically was created to deal with name clashes when making public.  However, now when you make public, you can change the name of the new public collection you are creating, so it is a bit redundant.  Maybe when you make public into an existing collection, can still be useful.  But, I think removing this may be fine too.

PR #1108, opened 2020-01-04
Google oauth integration

radmen:
Hey, I'm a contract developer working for Lygos. This PR introduces a feature that we discussed during our recent call - integration with Google OAuth.

Please mind that for now, it's a draft PR, which will be updated in the upcoming days. Feel free to leave comments.

I'll try to update this PR description as my work progress.

---

### Dependencies
- added:
  - `joi` - used to validate schemas (or in general value types)
  - `passport` - library which provides various auth strategies
  - `crypto-random-string` - secure random string generator
  - `connect-flash` - support for flash messages

### Added
- external auth provider section on the setup page
- "Sign in with Google" button on the login page
- "Connect  with Google" button on the profile page
- `plainOrHtml()` helper function in response object
- `UserExternalProfile` database model
- `utils.getUserIdFromToken()`
- ability to create an account via OAuth provider
- ability to connect existing profile with OAuth provider

### Changed
- moved app middleware to `app-middleware.js` module
- refactored multiple modules

### Authorization overview

- the `passport` module initializes Passport library
- retrieved profile data will not be serialized in any way - it will be passed to the callback endpoint which will handle everything
- there are two registered routes:
  - `/auth/external` which redirects to OAuth provider login page
  - callback endpoint (it is defined during the setup) which receives profile data and authenticates the user

### `plainOrHtml()`
I've noticed that many views try to support a case when the client can't accept HTML responses. The code used for this was copy-pasted in multiple places. I find this to be error-prone and hard to maintain, so I decided to encapsulate this behavior and created the `plainOrHtml()` helper function, which is attached to the response object.

It can be used in the following way:

```js
(req, res) => {
  res.plainOrHtml({
    { status: 201, message: 'Created' },
    () => {
      res.send(renderTemplate())
    }
  })
}
```

The first argument defines HTTP status code and message which should be returned when the client doesn't accept `text/html`. Otherwise, execute the callback. The callback should operate in `res` in a way that it will return any response to the client.

The callback not always returns an HTML response (hence I wanted initially to name it `plainOrHttp`) yet I thought that it clearly describes the intention behind, and it shouldn't generate any confusion.

### Installation

**Pre-requisites**
- set up [Google OAuth 2.0 keys](https://support.google.com/cloud/answer/6158849?hl=en)
- make sure that the Redirect URI points to the application (don't forget to add the path to the address). Example: `http://sbh.com/auth/google/callback`. URIs to `localhost` are invalid and will be rejected by Google.

**On on running instances**
- install new dependencies with `yarn`
- add to `config.local.json` following section:
```json
{ 
  "externalAuth": {
    "enabled": true,
    "provider": "google",
    "google": {
      "clientId": "XXX.apps.googleusercontent.com",
      "clientSecret": "rWIWoT7GJwqXXXXXXX",
      "callbackUrl": "http://sbh.com/auth/google/callback"
    }
  }
}
```
- install auth strategy with `yarn add passport-google-oauth@~2.0.0`
- restart the server

**During the setup**
- follow the setup page
- in the "Set up external auth provider" select "Google" and fill the following fields

---

### TODO:
- [x] flash an error message when failed to create a new user
- [x] update docs
- [x] install required strategy package during the setup


radmen:
Whenever it's possible I try to follow the [Boy Scout Rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html). This means that I'll try to refactor some parts of the modules I'm working on.

SynBioHub mixes various coding styles which means that my work might add yet another style. Since we agreed to use Node v11 I can start using things like:
- `async`/`await`
- short closures
- object and array destructuring

@3ach are you folks ok with that, or should I rather keep the existing coding style? As a reference, you can check [views/setup.js](https://github.com/radmen/synbiohub/blob/38de934971182a7aa9c198536391a42442c9d210/lib/views/setup.js) which I already refactored.

cjmyers:
Understand that your style may differ from ours.  The code has evolved beginning with callbacks, and no promises.  I've heard of async/await, but I've not learned how to use it as of yet.  While I'm not against you refactoring code, please be aware that it may slow down approval process, since it will take a bit more effort to validate that no harm has been done in the refactor.

Note that your updates may change test results, see instructions here for updating them:

lib/views/setup.js
```diff
@@ -34,155 +67,125 @@ module.exports = function (req, res) {
...
-
-    errors.push(err.stack)
-
-    return setupForm(req, res, settings, {
-      errors: errors
+  } catch (err) {
+    return setupForm(req, res, updatedSettings, {
+      errors: [
+        'Could not create user',
+        err.stack
```
3ach: Should the stack be spread?

radmen: No. Prior to this changes to error was pushed using following code:

```js
     errors.push('Could not create user')
     errors.push(err.stack) 
```

radmen:
@3ach @cjmyers Folks, I think this MR is ready for review.

lib/passport/util.js
```diff
@@ -0,0 +1,29 @@
+const { exec } = require('child_process')
+
+const dependencies = {
```
radmen: This could be moved to `package.json`, somewhere below the dependencies list. I think it would make it easier to find those additional dependencies.

lib/util.js
```diff
@@ -1,7 +1,28 @@
+const config = require('./config')
 
-var sha1 = require('sha1')
-var config = require('./config')
+const createTriplestoreID = (username) => `user/${username}`
```
radmen: check if this method is used somewhere else.

cjmyers:
I've updated the test results and committed these changes.  All the updated results looked fine to me.  How can I test your new feature?

radmen:
@cjmyers I've updated the PR description. At the bottom of it, you'll find instruction on how to start using OAuth in SBH. I hope that this is clear, if not let me know.

cjmyers:
Thanks.  I'm swamped hosting a meeting this week, but I will try it out next week.

nmih:
Just a note for those trying this out -- I attempted to set this up using the docker-compose installation method and it worked, but only after a restart of the docker container. The full steps I took were:
1. Run the docker-compose installation method using this PR
1. Set up SynBioHub with the OAuth keys
1. Restart the Docker containers -- if I didn't, I was getting a 500 not found error when clicking the log in with google button that directs to /auth/external
1. Log in with Google

Also, I'm wondering if there should be an intermediate option between allowing public account creation and not -- on our side, we'd only like users to login with their google account. But since that means public account creation is still required, I propose a third option that simply disables the "I need an account" button on the login page, while still exposing the oauth login option.

lib/views/profile.js
```diff
@@ -1,36 +1,46 @@
...
   }
 }
 
-function update (req, res) {
-  db.model.User.findById(req.user.id).then(user => {
-    if (req.body.name) { user.name = req.body.name }
-
-    if (req.body.affiliation) { user.affiliation = req.body.affiliation }
+function hasPasswordChanged ({ password1, password2 }) {
+  return !!(password1 && password2 && password1 === password2)
```
3ach: This is difficult to read, is there a reason to make it so terse?

lib/createUser.js
```diff
@@ -1,48 +1,35 @@
...
       username: info.username,
       affiliation: info.affiliation,
-      password: sha1(config.get('passwordSalt') + sha1(info.password)),
-      graphUri: graphUri,
-      isAdmin: info.isAdmin !== undefined ? info.isAdmin : false,
-      isCurator: info.isCurator !== undefined ? info.isCurator : false,
-      isMember: info.isMember !== undefined ? info.isMember : false
+      password: User.hashPassword(info.password),
+      graphUri,
+      isAdmin: !!info.isAdmin,
```
3ach: This is clever but a little difficult to parse if you are not familiar with how booleans should resolve. I think I prefer the old-style here.

lib/views/profile.js
radmen: Yes - I want to be sure that `hasPasswordChanged` returns boolean.

JS has this ugly behavior that when one chain expressions with `&&`, it will return either the most-right expression (considering that the preceding expressions are truthy) or a falsey value (which can be everything - empty string, `null`, `undefined` etc.). Double negation is used to make sure that the function will return a boolean.

lib/createUser.js
radmen: It's a common pattern used in JS applications. It guarantees that a variable will be converted to a boolean value.

I can revert the changes to the previous format if you want to.

lib/views/profile.js
cjmyers: I think it is okay to keep, but could you add a comment.

lib/createUser.js
cjmyers: Ditto.  Please add a comment about the !! usage.  Took me a bit to sort out what you were trying to do.  Also, "graphUri," is a new one for me.  I'm assuming it assumes:  "graphUri: graphUri".  I think I prefer explicit here.

radmen: @3ach @cjmyers I've made the code more explicit in 3eeeaf4

lib/views/profile.js
radmen: @3ach @cjmyers I've made the code more explicit in 3eeeaf4

cjmyers:
@nmih I'm fine merging this code now.  It passes our tests, so I'm reasonably confident that it is safe to do so.  However, I don't have Google credentials for testing the new feature.  Have you tested it?  Is it working as you expected?

lib/createUser.js
cjmyers: Thanks

nmih:
@cjmyers Yes, I've been able to test it out and everything seems to be working as intended. Merging would be excellent.

udp:
This all looks good to me. Nice work!

I think the conversion of the entire codebase to async/await is inevitable (I did this for [synbiohub lab](https://github.com/synbiohub/synbiohub-lab])), so using it for any new code seems like a good idea.

cjmyers:
@udp in order to run our testing, we need to move this fork into a local branch.  Do you know how to do this?  Once it runs through our testing, we can merge this PR.

udp:
@cjmyers sure, I'll do it now

udp:
https://github.com/SynBioHub/synbiohub/tree/google-oauth-integration

cjmyers:
How do I generate a pull request for that branch?

udp:
New pull request, then set the "compare" branch to google-oauth-integration.

cjmyers:
Ok, created.  Will need another approval, since I cannot approve the PR I create.

cjmyers:
We opened a new PR #1138, which has been merged into master.

PR #1128, opened 2020-01-13
Refactor plugins/streams to work with evaluate

### LINKED ISSUE #842: Add support for download plugins
Examples of download plugins include:

1) SnapServer generated SnapGene file and/or plasmid map.
2) Visualizations in other formats than PNG.

### LINKED ISSUE #1075: Change download plugins to use evaluate endpoint
When a page is loaded, each download plugin status is checked, then it is sent the type of the object via the evaluate endpoint for that plugin.  Based on the response, it will add a button to the page if the plugin can serve that type of object.  When the button is selected, then the run endpoint is called and the file return is downloaded to the user's system.

3ach:
This makes it so that plugins are not run until they are needed, and calls evaluate to determine whether or not the plugin box/link shoul be removed. 

Closes #1075 
Closes #842


cjmyers:
This look fine, but will it require the iGEM and Viz plugins to be updated?

Issue #1126, opened 2020-01-12
Submit endpoint seems to hang when overwrite_merge==3


bbartley:
The submission endpoint called by pySBOL was changed from `remoteSubmit` to `submit`.  Now, when running integration tests between pysbol/synbiohub, overwrite_merge modes 0, 1, and 2 execute successfully but overwrite_merge==3 seems to hang.

cjmyers:
I'm unable to reproduce this using either the browser or libSBOLj.  Can you provide more details about the test you are running?

bbartley:
The test in simple terms is as follows.  Create a `ComponentDefinition` and assign a `role` as an SO term.  Push this to a `Collection`.  Now change the `role` to a new SO term and attempt to merge and overwrite the `Collection`.

Here is the longform Python code:
```
from sbol import *
from urllib3.exceptions import HTTPError

RESOURCE = 'https://hub-staging.sd2e.org'
SPOOFED_RESOURCE = 'https://hub.sd2e.org'
USER = <user>
PASSWORD = <password>
PART_SHOP = PartShop(RESOURCE)
if SPOOFED_RESOURCE:
    PART_SHOP.spoof(SPOOFED_RESOURCE)
PART_SHOP.login(USER, PASSWORD)

TEST_COLLECTION = 'pySBOL_test'
TEST_COLLECTION_URI = RESOURCE + '/user/' + USER + '/' + TEST_COLLECTION + '/' + TEST_COLLECTION + '_collection/1'
Config.setOption('sbol_typed_uris', False)

# test submit
print('Testing submit...', end = '')
doc = Document()
doc.displayId = TEST_COLLECTION
doc.name = 'pySBOL test'
doc.description = 'A temporary collection used for running pySBOL integration tests'
doc.componentDefinitions.create('foo')
PART_SHOP.submit(doc)
print('SUCCESS')

# test submit -- overwrite
print('Testing submit overwrite...', end = '')
doc = Document()
doc.displayId = TEST_COLLECTION
doc.name = 'pySBOL test'
doc.description = 'A temporary collection used for running pySBOL integration tests'
bar = doc.componentDefinitions.create('bar')
bar.roles = SO_PROMOTER
PART_SHOP.submit(doc, TEST_COLLECTION_URI, 1)
print('SUCCESS')

# test submit -- merge and prevent
print('Testing merge and prevent...', end = '')
doc = Document()
doc.displayId = TEST_COLLECTION
[...]

PR #1129, opened 2020-01-14
Addresses merge bug #1126

cjmyers:


cjmyers:
Fixes #1126

lib/views/submit.js
```diff
@@ -349,7 +349,7 @@ function submitPost (req, res) {
           }
           res.send(pug.renderFile('templates/views/errors/errors.jade', locals))
         }
-      })
+      }) */
```
udp: How come this is commented out?

cjmyers: It seems to be causing an unhandled promise rejection.  Actually, I'm still seeing unhandled promise rejections when you try to merge and are prevented due to trying to submit non-identical objects into an existing collection, but this seemed to move it closer to where it was happening.  It is strange. Still bad at promises

udp: Refactoring to use async/await should make it much more clear where exceptions are going.  Will accept to solve the immediate issue but we need to sort this soon.

cjmyers:
Agreed.

Issue #1077, opened 2019-12-13
Refactor submit.js to fix Heisenbug.


cjmyers:
There is a timing issue at line 105 on submit.js.  This appears occasionally when running the SBH SBOLTestSuite.  It appears to occur consistently when adding a collection with the new auth module.  

Refactor the code to see if we can eliminate this timing issue on processing of forms, and test both in master and the auth PR.

3ach:
I'm pretty sure this can be closed. We've been running on the new submit for ~3 months.

Issue #1137, opened 2020-01-15
Refactor to use async/await instead of promises


udp:
Examples in https://github.com/synBioHub/synbiohub-lab

Issue #1130, opened 2020-01-14
Unhandled promise rejection in submit.js


cjmyers:
UnhandledPromiseRejectionWarning: TypeError: Cannot destructure property `success` of 'undefined' or 'null'.
    at saveTempFile.then.then.then (/Users/myers/git/synbiohub/lib/views/submit.js:437:17)
    at process.internalTickCallback (internal/process/next_tick.js:77:7)
error: (node:75744) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 5)

This happens when you try to merge into existing collection an object that would overwrite an existing item (overwrite_merge == 2).  Submit is blocked as it should be, so it does not cause a noticeable problem, but it should be removed.

cjmyers:
Fixed by PR #1157

PR #1157, opened 2020-02-19
Refactor submit

3ach:
A refactor of submit as a first pass at addressing #1130. Also part of #1137.


3ach:
Closes #1077

3ach:
Looks like your changes fixed the issues, the test suite is now (actually) passing

cjmyers:
Submitting into an existing collection does not seem to work.

PR #1165, opened 2020-03-03
Fix issue with merge when there is not found exception

cjmyers:


java/src/main/java/org/synbiohub/PrepareSubmissionJob.java
```diff
@@ -482,8 +484,14 @@ public void visit(Identified identified, TopLevel topLevel) {
 						SBOLDocument tlDoc;
 						try {
 							tlDoc = sbh.getSBOL(URI.create(topLevelUri));
-						} catch (SynBioHubException e) {
-							tlDoc = null;
+						} catch (SynBioHubException e1) {
+							StringWriter sw = new StringWriter();
```
3ach: You can't `printStackTrace(System.err)` ?

java/src/main/java/org/synbiohub/PrepareSubmissionJob.java
```diff
@@ -482,8 +484,14 @@ public void visit(Identified identified, TopLevel topLevel) {
 						SBOLDocument tlDoc;
 						try {
 							tlDoc = sbh.getSBOL(URI.create(topLevelUri));
-						} catch (SynBioHubException e) {
-							tlDoc = null;
+						} catch (SynBioHubException e1) {
```
3ach: Won't this still fail in the event of a NotFoundException?

cjmyers: I changed libSBOLj to return null on NotFoundException.

java/src/main/java/org/synbiohub/PrepareSubmissionJob.java
cjmyers: Probably, but I was also converting to a string to return in errorLog.

PR #1170, opened 2020-03-04
fix order to fix share

### LINKED ISSUE #1169: Share links do not appear to be working
Error: getType: https://synbiohub.org/user/myers/Nic/Nic_collection/1/af00936caf838ff71ebe45d404aad886b96007ce/share not found
  at sparql.queryJson.then (/Users/myers/git/synbiohub/lib/query/local/type.js:12:29)
  at process.internalTickCallback (internal/process/next_tick.js:77:7)

cjmyers:
Fixes #1169

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -373,14 +373,14 @@ function App () {
   app.get('/public/:collectionId/:displayId', views.persistentIdentity)
   app.get('/user/:userId/:collectionId/:displayId', views.persistentIdentity)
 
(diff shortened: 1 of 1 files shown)


cjmyers:
Yeah, the issue is the app.js is matched from top to bottom.  The "*" in the match causes the match to terminate and not hit the share ones below.  They need to be above the "non-share" ones.

PR #1172, opened 2020-03-18
Fixes #1167

### LINKED ISSUE #1167: Incorrect login token causes an unhandled promise rejection

cjmyers:


cjmyers:
Fixed problems with logout via the API.  It was not actually logging out via the API, and it also was not checking correctly if one is logged out before sending along.

lib/apiTokens.js
```diff
@@ -11,6 +11,9 @@ function createToken (user) {
 
   return token
 }
+function deleteToken (token) {
+  tokens[token] = null
```
3ach: should be `delete tokens[token]`

PR #1194, opened 2020-04-02
Added sequence search page

yu-eric:
A separate page for sequence search has been added.


lib/search.js
```diff
@@ -138,6 +138,10 @@ search.lucene = function lucene (value) {
       criteriaStr += '   FILTER (xsd:dateTime(?mdate) <= "' + query[1] + 'T23:59:59Z"^^xsd:dateTime) '
     } else if (query[0] === 'modifiedAfter') {
       criteriaStr += '   FILTER (xsd:dateTime(?mdate) >= "' + query[1] + 'T00:00:00Z"^^xsd:dateTime) '
+    } else if (query[0] === 'globalsequence') {
+      return `?subject sbol2:sequence ?seq .\n    ?seq sbol2:elements "${query[1]}" .`
+    } else if (query[0] === 'sequence') {
+      return `?subject sbol2:sequence ?seq .\n    ?seq sbol2:elements "EXACT${query[1]}" .`
```
3ach, line 144: Should there be a space after "EXACT" ?

lib/views/sbsearch.js
```diff
@@ -0,0 +1,62 @@
...
+  var locals = {
+    config: config.get(),
+    section: 'admin',
+    adminSection: 'explorer',
+    user: req.user
+  }
+  res.send(pug.renderFile('templates/views/sbsearch.jade', locals))
+}
+
+function post (req, res) {
```
3ach, line 26: If you exported an async function, you can make this async too

lib/views/sbsearch.js
```diff
@@ -0,0 +1,62 @@
...
+    flags['exactSearch'] = true
+  }
+
+  if (req.body.maxAccepts.length > 0) {
+    flags['maxAccepts'] = req.body.maxAccepts
+  }
+  if (req.body.id.length > 0) {
+    flags['id'] = req.body.id
+  }
+  searchText(searchSequence, flags).then(() => {
```
3ach, line 42: Then `await` this result.

cjmyers:
This looks good to me.  One minor point is using the sequence search page, it fills the search bar with:

sequence=advancedsequencesearch&

Can we fill it with the necessary search info, so if you press search again, it will work.

Also, does max accept work for exact?  I increased to 50 for global, which worked, but I still got 10 when I tried exact.

cjmyers:
Disregard last part.  There were only 10 exact matches.  I tried another example, and it returned them all.  This looks good to me.  I approve too.

PR #1216, opened 2020-04-19
Make user endpoints API friendly

### LINKED ISSUE #1148: Make User endpoints API friendly
These endpoints aren't user friendly: 
app.all('/logout', views.logout)
app.all('/resetPassword/token/:token', actions.resetPassword)
app.all('/resetPassword', views.resetPassword)
app.post('/setNewPassword', actions.setNewPassword)
app.all('/profile', requireUser, views.profile)

cjmyers:
Closes #1148


lib/views/profile.js
```diff
@@ -48,13 +53,18 @@ function display (req, res) {
...
-    section: 'profile',
+  var locals = {
     user,
     externalAuthProvider: dotget(config.get('externalAuth'), 'provider'),
     externalProfiles
   }
-
-  res.send(pug.renderFile('templates/views/profile.jade', locals))
+  if (!req.accepts('text/html')) {
+    return res.status(200).header('content-type', 'application/json').send(locals)
```
3ach, line 62: This locals is okay

cjmyers: This is sending back the user and any external (OAuth) information.  Are you think I should send only back a subset?  I was curious about this.  It does send back the encrypted password currently.  Is that a concern?

3ach: Hmm, doesn't sound great to send back, but I don't know off the top of my head

cjmyers:
Removed password from return value

PR #1223, opened 2020-04-25
This fixes the problem with unhandled promise rejection

### LINKED ISSUE #1212: Delete user with invalid id causes promise exception
curl -X POST -H "Accept: text/plain" -H "X-authorization: ef82f692-c35c-4761-8b74-049f2cc02e01" -d "id=2" http://localhost:7777/admin/deleteUser

Unhandled rejection Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:470:11)
    at ServerResponse.header (/Users/myers/git/synbiohub/node_modules/express/lib/response.js:771:10)
    at ServerResponse.send (/Users/myers/git/synbiohub/node_modules/express/lib/response.js:170:12)
    at Model.db.model.User.findById.then.then.catch (/Users/myers/git/synbiohub/lib/actions/admin/deleteUser.js:14:28)

cjmyers:
Closes #1212

### DIFF
--- a/lib/actions/admin/updateUser.js
+++ b/lib/actions/admin/updateUser.js
@@ -11,8 +11,9 @@ module.exports = function (req, res) {
       user.isMember = !!req.body.isMember
       user.isCurator = !!req.body.isCurator
       user.isAdmin = !!req.body.isAdmin
-      user.save()
-      res.status(200).send('User (' + req.body.id + ') updated')
+      user.save().then(() => {
+        res.status(200).send('User (' + req.body.id + ') updated')
+      })
     } else {
       res.status(400).send('Must provide a valid user id')
     }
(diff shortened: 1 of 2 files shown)


cjmyers:
Note from now on I'm adding all as reviewers, but only review is required.  Please review if you get this and you are comfortable doing so.

cjmyers:
@3ach Committed requested changes, please check.

lib/actions/admin/updateUser.js
```diff
@@ -11,12 +11,11 @@ module.exports = function (req, res) {
       user.isMember = !!req.body.isMember
       user.isCurator = !!req.body.isCurator
       user.isAdmin = !!req.body.isAdmin
-      return user.save()
+      user.save()
```
3ach, line 14: Sorry, I missed something. It should be `user.save.then(() => { res.... })` for the 200, so that it will only send a 200 if the save succeeds, otherwise it will catch the error and send the 500. The way it is now, it will send the 200 without waiting for the `destroy` or `save` to complete, and if they have an error it will throw the error about being unable to set headers.

PR #1224, opened 2020-04-25
Fixes promise rejection when SendGrid returns an error

### LINKED ISSUE #1214: Unhandled promise rejection error when bad request to SendGrid
When doing something that causes a request to SendGrid, such as adding a new user, there is an unhandled promise rejection when the SendGrid configuration is invalid.

error: (node:40386) UnhandledPromiseRejectionWarning: SendGridError: Response error
    at /Users/myers/git/synbiohub/node_modules/sendgrid/lib/sendgrid.js:123:19
    at IncomingMessage.<anonymous> (/Users/myers/git/synbiohub/node_modules/sendgrid-rest/lib/client.js:112:9)
    at IncomingMessage.emit (events.js:187:15)
    at endReadableNT (_stream_readable.js:1094:12)
    at process.internalTickCallback (internal/process/next_tick.js:72:19)
error: (node:40386) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)

cjmyers:
Closes #1214 
This happens, for example, when there is no valid API key specified

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,4 +1,8 @@
 # GitHub Actions
+## Test
+This is the workflow described in the file `test.yml`.
+It runs the SBH test suite, the SynBioHub test suite, and then if both succeed the image is pushed to Docker Hub.
+
 ## Release
 This is the workflow described in the file `release.yml`.
 It is composed of two jobs: `retag-release` and `update-compose`.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,88 @@
+name: Test and publish snapshot
+
+on: push
+
+jobs:
+  build: 
+    name: Build Docker image
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build the Docker image
+      run: |
+        docker build . --file docker/Dockerfile --tag synbiohub/synbiohub:snapshot-standalone
+    - name: Package image
+      run: |
+        docker save synbiohub/synbiohub:snapshot-standalone | gzip > sbh.tar.gz
+    - uses: actions/upload-artifact@v2
+      with:
+        name: sbh-image
+        path: sbh.tar.gz
+  sboltests:
+    name: SBOL Test Suite
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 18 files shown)


cjmyers:
Closes #1214

lib/views/resetPassword.js
```diff
@@ -56,17 +56,47 @@ function resetPasswordPost (req, res) {
...
+    if (err === {}) {
+      if (!req.accepts('text/html')) {
+        return res.status(200).header('content-type', 'text/plain').send('If a user was found associated with the specified e-mail address, a link to reset your password has been sent.')
+      } else {
+        res.send(pug.renderFile('templates/views/resetPasswordDone.jade', locals))
+      }
     } else {
-      res.send(pug.renderFile('templates/views/resetPasswordDone.jade', locals))
+      var error = err
+      var statusCode = 500
```
3ach, line 75: I think you can just do `statusCode = err.response.statusCode || 500`

lib/views/resetPassword.js
```diff
@@ -56,17 +56,47 @@ function resetPasswordPost (req, res) {
...
+      if (err.message) {
```
3ach, line 76: Do you only have `err.response` when you have `err.message`?

lib/views/resetPassword.js
cjmyers: What if err is undefined?  Would this not create an exception?  The logic is complex because I was not entirely sure what I would be getting back.  This is a error returned by Sendgrid.  I was being very particular, since I was parsing the JSON coming back, and I think assuming and dereferencing can lead to exceptions.  I'm not sure though about the syntax you give though if that is more robust to dereference problems.

lib/views/resetPassword.js
cjmyers: I'm not sure.  I just looked at what Sendgrid sent back, and I parsed it out.  However, I was nervous that I could not rely on the JSON structure to be fixed.

lib/views/resetPassword.js
3ach: I just tested, and it will create an exception. Can be left as-is.

PR #1225, opened 2020-04-26
Add support to remove a member of a collection

### LINKED ISSUE #809: Allow deletion from a collection
When a collection member gets deleted but somehow not deleted from the collection, it becomes a "ghost" element on SynBioHub that can't be deleted.  Can we allow one to delete at the collection level, not just when looking at the individual part?

### LINKED ISSUE #1214: Unhandled promise rejection error when bad request to SendGrid

cjmyers:
Closes #809

### DIFF
--- a/lib/mail/resetPassword.js
+++ b/lib/mail/resetPassword.js
@@ -4,7 +4,7 @@ const loadTemplate = require('../loadTemplate')
 const config = require('../config')
 
 function sendResetPasswordMail (user) {
-  sendMail(user, 'Reset your password', loadTemplate('mail/resetPassword.txt', {
+  return sendMail(user, 'Reset your password', loadTemplate('mail/resetPassword.txt', {
 
     link: config.get('instanceUrl') + 'resetPassword/token/' + user.resetPasswordLink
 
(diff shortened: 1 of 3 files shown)


cjmyers:
This PR modifies the deletion from collection view functionality as follows:

1) If the selected object was originally submitted as part of this collection (i.e., it has same URI prefix as the private collection being viewed), then there is a trash can which when selected will remove the object entirely.

2) If the selected object was not originally submitted as part of this collection (i.e. it does not have the same URI prefix as the private collection being viewed), then there is an unlink icon which when selected will remove this object from the collection.

cjmyers:
PR has wrong files closing.

PR #1226, opened 2020-04-26
Add support to remove a member of a collection

### LINKED ISSUE #809: Allow deletion from a collection

cjmyers:
Closes #809


cjmyers:
This PR modifies the deletion from collection view functionality as follows:

If the selected object was originally submitted as part of this collection (i.e., it has same URI prefix as the private collection being viewed), then there is a trash can which when selected will remove the object entirely.

If the selected object was not originally submitted as part of this collection (i.e. it does not have the same URI prefix as the private collection being viewed), then there is an unlink icon which when selected will remove this object from the collection.

lib/actions/removeMembership.js
```diff
@@ -0,0 +1,70 @@
...
+  var templateParams = {
+    uri: uri,
+    memberUri: memberUri
+  }
+
+  var removeQuery = loadTemplate('sparql/removeMembership.sparql', templateParams)
+
+  console.log(removeQuery)
+
+  return getOwnedBy(uri, graphUri).then((ownedBy) => {
```
3ach, line 30: Same comment as add -- can we check ownership earlier?

cjmyers: Not sure where you have in mind.  The parts above are just defining the removeQuery.

PR #1227, opened 2020-04-26
Add membership to collections

### LINKED ISSUE #1110: Add to collection membership
Should be able to add an object as a member of a collection.  This should be able to both be done upon submission and also updated after submission both from browser and API.

cjmyers:
Closes #1110


lib/app.js
```diff
@@ -283,8 +284,12 @@ function App () {
   app.post('/public/:collectionId/:displayId/:version/icon', requireUser, uploadToMemory.single('collectionIcon'), actions.updateCollectionIcon)
 
   app.get('/public/:collectionId/:displayId/:version/removeCollection', requireAdmin, actions.removeCollection)
-  app.get('/user/:userId/:collectionId/:displayId/:version/removeCollection', requireUser, actions.removeCollection)
+  app.get('/public/:collectionId/:displayId/:version/addToCollection', requireAdmin, views.addToCollection)
+  app.post('/public/:collectionId/:displayId/:version/addToCollection', requireAdmin, views.addToCollection)
 
+  app.get('/user/:userId/:collectionId/:displayId/:version/removeCollection', requireUser, actions.removeCollection)
+  app.get('/user/:userId/:collectionId/:displayId/:version/addToCollection', requireUser, views.addToCollection)
+  app.post('/user/:userId/:collectionId/:displayId/:version/addToCollection', requireUser, views.addToCollection)
```
3ach, line 292: Do they need share endpoints?

lib/views/addToCollection.js
```diff
@@ -0,0 +1,119 @@
...
+const pug = require('pug')
+const config = require('../config')
+const sparql = require('../sparql/sparql')
+const getOwnedBy = require('../query/ownedBy')
+
+module.exports = function (req, res) {
+  req.setTimeout(0) // no timeout
+
+  function addToCollectionsForm (req, res) {
+    var collectionQuery = 'PREFIX dcterms: <http://purl.org/dc/terms/> PREFIX sbol2: <http://sbols.org/v2#> SELECT ?object ?name WHERE { ?object a sbol2:Collection . OPTIONAL { ?object dcterms:title ?name . }}'
```
3ach, line 11: Can we put this in a .sparql file?

lib/views/addToCollection.js
```diff
@@ -0,0 +1,119 @@
...
+    function sortByNames (a, b) {
+      if (a.name < b.name) {
+        return -1
+      } else {
+        return 1
+      }
+    }
+
+    return sparql.queryJson(collectionQuery, req.user.graphUri).then((collections) => {
+      collections.forEach((result) => {
```
3ach, line 22: A `.map()` might make more sense here than `forEach`.

lib/views/addToCollection.js
```diff
@@ -0,0 +1,119 @@
...
+    var templateParams = {
+      uri: uri,
+      memberUri: memberUri
+    }
+
+    var addQuery = loadTemplate('sparql/addMembership.sparql', templateParams)
+
+    console.log(addQuery)
+
+    return getOwnedBy(uri, req.user.graphUri).then((ownedBy) => {
```
3ach, line 76: Can we check ownedBy earlier? I think that makes more sense to read rather than interrupting the process of adding.

lib/views/sbsearch.js
```diff
@@ -3,6 +3,10 @@ var pug = require('pug')
 const config = require('../config')
 
 module.exports = async function (req, res) {
+  res.setHeader('Cache-Control', 'no-cache')
+  res.setHeader('Cache-Control', 'no-store')
+  res.setHeader('Pragma', 'no-cache')
```
3ach: What're these about?

templates/layouts/topLevel.jade
```diff
@@ -60,6 +60,10 @@ block generalTopLevelButtons
                         li.btn.btn-success
                                 a(href=meta.url + '/remove',title='Remove',style="color:white", onclick="return confirm('Are you sure you want to delete this?')")
                                     span.fa.fa-remove
+                if locals.user
```
3ach, line 63: probably also `canEdit`, right?

3ach:
Also the branching was a little weird, pulled in some commits from other PRs. I rebased them out but good to check next time.

lib/app.js
cjmyers: I was not sure if it made sense to allow share endpoints to have this capability or not.  I thought share endpoints were basically read-only, and add owner was used to give edit capability.

lib/views/addToCollection.js
cjmyers: Actually it is there (getCollections.sparql).  Will switch to use it.

lib/views/addToCollection.js
cjmyers: Ok, I switched, and it works.  Not sure I understand the difference here.  I think we are using this same structure in other places.

lib/views/addToCollection.js
cjmyers: Not sure where you have in mind.  There is not actual code earlier, since much of the code before is just a function for the form.

lib/views/sbsearch.js
cjmyers: This is fixing a browser bug for sequence-search.  Was not supposed to be part of this PR.

templates/layouts/topLevel.jade
cjmyers: No, because you are not actually editing this object, you are copying it into your own collection.

PR #1235, opened 2020-05-05
Removed ability for user to change the default public graph

yu-eric:


cjmyers:
Looks good, but please double check that you have also eliminated this from any code that sends the config to SBOLExplorer.

PR #1237, opened 2020-05-10
Fixes several problems with add/edit/remove fields

### LINKED ISSUE #1036: Adding a type is not working
When trying to add a type to an object that already has a type, it is not getting added.

### LINKED ISSUE #1109: Cannot remove title/description
Should be able to remove title/description by entering a blank one.  

For title, if removed it should show the displayId.

Note that if you edit the title, and there is none, you will end up with the id getting copied to the title (name).

cjmyers:
Closes #1109 and 
Closes #1036

### DIFF
--- a/browser/field-editor.js
+++ b/browser/field-editor.js
@@ -36,9 +36,7 @@ function save($input, $elem, toEdit, oldVal) {
     appendEditor(0, $elem)
 
     updateSparql(newVal, oldVal, toEdit, (newText) => {
-        $elem.text(newText)
-        $input.replaceWith($elem)
-        appendEditor(0, $elem)
+        location.reload()
     })
 }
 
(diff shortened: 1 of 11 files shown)


templates/layouts/identified.jade
```diff
@@ -103,11 +106,12 @@ block content
                         a(href='/search/modifiedBefore='+meta.modified.description.split(' ')[0]+'&modifiedAfter='+meta.modified.description.split(' ')[0]+'&',title='Find all records modified on this date') 
                             span.fa.fa-search
 
-                if meta.description && meta.description != ''
+                if (meta.description && meta.description != '') || meta.canUpdate
```
3ach, line 109: Why not `&& meta.canUpdate` ?

tests/previousresults/getrequest_public-:collectionId-:displayId_bbapublic.html
```diff
@@ -178,7 +178,7 @@ <h5>
      <h4>
       LuxR protein generator
       <br/>
-      <a href="/search/?q=LuxR protein generator&lt;br/&gt;
+      <a href="/search/?q=LuxR protein generator
```
3ach, line 181: How did this change?

tests/previousresults/getrequest_public-:collectionId-:displayId_docker_public.html
```diff
@@ -221,7 +221,7 @@ <h5>
      <h4>
       testdescription0
       <br/>
-      <a href="/search/?q=testdescription0&lt;br/&gt;
+      <a href="/search/?q=testdescription0
```
3ach, line 224: How did this change?

templates/layouts/identified.jade
```diff
@@ -103,11 +106,12 @@ block content
...
 
-                if meta.description && meta.description != ''
+                if (meta.description && meta.description != '') || meta.canUpdate 
                     h4(class=meta.canUpdate ? "edit edit-description" : "")
-                        != meta.description 
-                        a(href='/search/?q='+meta.description,title='Find all records with terms in common with this description')
-                            span.fa.fa-search
+                        != meta.description
+                        if meta.description && meta.description != ''
+                            a(href='/search/?q='+meta.description.replace('<br/>',' '),title='Find all records with terms in common with this description')
```
3ach, line 113: Is this supposed to be part of this PR?

lib/api/removeObject.js
```diff
@@ -31,8 +34,8 @@ function tripleIsValid (subject, predicate, object) {
   return true
 }
 
-function formatObject (object) {
-  if (object.startsWith('http')) {
+function formatObject (field, object) {
+  if (field === 'wasDerivedFrom' || field === 'type' || field === 'role') {
```
3ach, line 38: Why change the way this was handled? Won't this lead to a big long list of field ifs?

browser/field-editor.js
```diff
@@ -170,16 +178,16 @@ function updateAddRemove() {
 
   // Remove existing adders and removers
   $("a[class|=add]").remove()
-  $("a[class|=remove]").remove()
+  //$("a[class|=remove]").remove()
```
3ach, line 181: Why aren't you removing the existing removers? Won't this lead to lots of removers getting added?

browser/field-editor.js
```diff
@@ -68,69 +66,79 @@ function getField(elem, idx) {
...
+  $elem.append(removeLink)
+
+  $('.' + removeClass).on('click', () => {
+    let $trashes = $(".remove-" + toRemove)
+    let text = $elem.attr("editText") || $elem.text()
+    
+    $trashes.remove()
+
+    removeSparql(text, toRemove, () => {
+      location.reload()
```
3ach, line 128: Why reload the page instead of just updating the field?

browser/field-editor.js
```diff
@@ -36,9 +36,7 @@ function save($input, $elem, toEdit, oldVal) {
     appendEditor(0, $elem)
 
     updateSparql(newVal, oldVal, toEdit, (newText) => {
-        $elem.text(newText)
-        $input.replaceWith($elem)
-        appendEditor(0, $elem)
+        location.reload()
```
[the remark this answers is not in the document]
cjmyers: There were a lot of problems in the pages getting messed up by the partial reloads.  This was the only way I could think to correct them all.

templates/layouts/identified.jade
cjmyers: This allows you to add a description even when there is not one initially.

tests/previousresults/getrequest_public-:collectionId-:displayId_bbapublic.html
cjmyers: There were extra br characters getting added to the description/

tests/previousresults/getrequest_public-:collectionId-:displayId_docker_public.html
cjmyers: ditto

templates/layouts/identified.jade
cjmyers: Yeah, this was the problem with extra breaks causing search by description to be broken.  I guess I could have put in separate PR, but I discovered as I was fixing these other issues.

lib/api/removeObject.js
cjmyers: URIs do not need to start with http, so technically this was not correct.  These three fields expect URIs, so whatever is entered is a URI.

browser/field-editor.js
cjmyers: I conditioned the other code to prevent them getting added.  Granted, this was a bit wonky since I really don't understand browser code well.  It is also likely why I reload pages all the time.

browser/field-editor.js
cjmyers: Pages were getting messed up for a variety of reasons.  This was the only way I could think to make sure everything got to the right state even when errors occur.

PR #1238, opened 2020-05-10
Fix bug with attachUrl when no type is selected

### LINKED ISSUE #1180: Attachurl broken
AttachURL crashes SynBioHub

cjmyers:
Closes #1180

### DIFF
--- a/lib/api/attachUrl.js
+++ b/lib/api/attachUrl.js
@@ -25,6 +25,14 @@ async function serve (req, res) {
     // error
   }
 
+  if (typeString === 'Pick an attachment type...') {
+    if (!req.accepts('text/html')) {
+      res.status(400).send('Must pick an attachment type')
+    } else {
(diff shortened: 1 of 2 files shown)


lib/api/attachUrl.js
```diff
@@ -25,6 +25,14 @@ async function serve (req, res) {
     // error
   }
 
+  if (typeString === 'Pick an attachment type...') {
+    if (!req.accepts('text/html')) {
+      res.status(400).send('Must pick an attachment type')
+    } else {
+      res.redirect(url)
```
3ach, line 32: Shouldn't it display an error on the page?

cjmyers: Agreed, but I was not sure how to do that.  Ideally, there should be popups with this one and the add/edit/remove fields ones when something goes wrong.  I settled on simply refreshing the page.

PR #1246, opened 2020-05-16
Add to coll bug

cjmyers:


cjmyers:
Found small bug need merged before demo tomorrow

PR #1256, opened 2020-05-25
Added health check for Virtuoso

### LINKED ISSUE #1140: Add virtuoso healthcheck
rootCollections now does not fail but returns empty when Virtuoso is down.  Therefore, we need a new Virtuoso healthcheck to check it directly.

cjmyers:
Closes #1140

### DIFF
--- a/lib/api/healthCheck.js
+++ b/lib/api/healthCheck.js
@@ -1,0 +1,17 @@
+const sparql = require('../sparql/sparql')
+
+function healthCheck (req, res) {
+  return sparql.query('ASK {}', null, 'text/plain').then(parseResult, handleError)
+
+  function parseResult (result) {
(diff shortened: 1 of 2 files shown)


jvscholz:
is this part of the api?

cjmyers:
Yes, once it is approved.  It is a pretty simple one.  It returns HTTP error code 200 when Virtuoso is up and running, and it returns HTTP error code 500 when it is not.

This raises an interesting point.  It would be good to document the reasons for various error codes for each endpoint.  I will log a separate issue.  I would put that at a lower priority though than testing.

jvscholz:
which category does it fall under?

cjmyers:
Probably admin

Chris

cjmyers:
Moved to /admin/virtuoso endpoint

PR #1260, opened 2020-05-26
Add curator information to ownership

### LINKED ISSUE #863: Request publication of data
we are planning to recommend SBH however one concern is whether users will be able to make their design public at the final stage of their projects?

3ach:
Adds (curator) to curators in the ownership share menu 

Closes #863


cjmyers:
Looks like testing needs to be updated on this one.

PR #1319, opened 2020-07-20
Tests new user

jvscholz:


cjmyers:
Tests says problem communicating with web-of-registries.

jvscholz:
updated

PR #1320, opened 2020-07-21
tests post updateUsersConfig

jvscholz:


jvscholz:
looks like this is failing because graphs is also comparing the created at timestamp and the updated at timestamp

cjmyers:
Yeah, you are correct on this.  We likely need a smart JSON comparison method that can ignore certain fields.

PR #1325, opened 2020-08-18
Edit auto indexing

yu-eric:
Now supported by explorer


yu-eric:
@jvscholz Can you check and see why the tests are failing? I don't think it's on my end, but I could be wrong.

PR #1328, opened 2020-08-26
fixes testing for some admin endpoints

jvscholz:


jvscholz:
looks like one of the endpoints is returning an error, fixing it right now

jvscholz:
update users config might fail because it seems to print a log of when it updated, might have to figure out a different testing solution

PR #1330, opened 2020-08-27
Remove explorer as it seems to be nondeterministic

jvscholz:
causes the test suite to fail on new trials


jvscholz:
you can merge #1331 instead of this branch cause that one has the same changes applied to it already

PR #1353, opened 2020-09-22
Visbol2 conform

benjhatch:
Fixed package-lock.json and updated to visbol 2.0


cjmyers:
Seems to still have the conflict.

PR #1384, opened 2020-11-10
Test add owner and remove owner (get and post)

jvscholz:


jvscholz:
The problems started happening after I added remove owner. Would that have caused such massive diffs? @cjmyers

jvscholz:
seems like these endpoints are nondeterminstic

PR #1392, opened 2020-11-24
tests addOwner, changes browse to GET only, tests GET endpoint for ed…

jvscholz:
…it and remove


cjmyers:
Seems to be failing tests.

cjmyers:
The failures look similar to the ones we saw before with different instance names, etc.  This is very strange.  You might try one more reset tests, but other than that I think you may need to look carefully at the order things are running.  This could indeed affect the size of the graphs with that test.  Not sure about the other issues.

PR #1405, opened 2021-01-14
Rearrange test suite to be sequential

jvscholz:


cjmyers (approved):

jvscholz:
@cjmyers 

looking at the log, I don't think it's timing out because of a dependency. if we compare the log from this pull request to a successful one, then you can see that chromium successfully downloads and extracts.

this or:
Screenshot from 2021-01-15 12-35-49

successful or:
Screenshot from 2021-01-15 12-36-04

it looks like it's waiting for the test suite to finish running- which never gets started in the first place for some reason. I created another pull request with an insignificant change to confirm that it's not related to the contents of the changes.

PR #1448, opened 2021-05-20
Remove sendgrid key

cjmyers:


cjmyers:
We cannot use a SendGrid key in our tests as it causes them to lock our sendgrid account.

benjhatch (approved):

PR #1452, opened 2021-06-16
Updated VisBol2

cjmyers:


benjhatch (approved):
Fingers crossed; still need to fix the testing

PR #1475, opened 2021-09-30
Update Dockerfile

JMante1:
Changed the base image to be more up to date based on NIST recommendations


JMante1:
for some reason the newer docker image can't install python using apk. Not sure what is going on

JMante1:
This is the command that is failing: apk add nss git openjdk8-jre openjdk8 maven python alpine-sdk libxml2-dev yarn g++ gcc bash raptor2 jq

cjmyers:
Looks like docker build is failing.

cjmyers:
Could one of those applications not be available for this new OS?

PR #1481, opened 2021-11-15
changed theme get to public

benjhatch:


benjhatch:
Changed theme get endpoint to be public (post endpoint is still private)

cjmyers (approved):

PR #1508, opened 2022-08-09
Zane dev plugins

zane-perry:
Added plugin endpoints to old backend

### DIFF
--- a/lib/plugins/pluginEndpoints
+++ b/lib/plugins/pluginEndpoints
@@ -6,85 +6,10 @@ const get = util.promisify(request.get)
 
 const config = require('../config')
 
-
-const findPlugin = (name) => {
-    const pluginList = config.get('plugins');
-
-
-    for (let plugin of pluginList.download) {
-        if(plugin.name === name) {
-            return [plugin.url, 'download'];
-        }
-    }
-
-    for (let plugin of pluginList.rendering) {
-        if(plugin.name === name) {
-            return [plugin.url, 'visual'];
-        }
-    }
-
-    for (let plugin of pluginList.submit) {
-        if(plugin.name === name) {
-            return [plugin.url, 'submit'];
-        }
-    }
-
-    return null;
-
-}
-
-const getStatus = (pluginUrl, res) => {
-    
-
-    return axios({
-        method: 'GET',
-        url: pluginUrl + 'status',
-        responseType: 'text'
-    }).then(response => {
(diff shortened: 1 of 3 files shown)


lib/plugins/pluginEndpoints
```diff
@@ -0,0 +1,117 @@
...
+            const data = decodeURIComponent(req.query.data); 
+            getEvaluate(pluginUrl, data, category, res);
+            break;
+        }
+        case 'run': {
+            const data = decodeURIComponent(req.query.data); 
+            getRun(pluginUrl, data, category, res);
+            break;
+        }
+        default: res.status(400).send('Error with request');
```
JMante1, line 37: I think this could be clearer: e.g. this plugin type is not known, try evaluate, status, run, or call instead

lib/plugins/pluginEndpoints
```diff
@@ -0,0 +1,117 @@
...
+
+module.exports = function (req, res) {
+
+    const name = req.query.name;
+    const endpoint = req.query.endpoint;
+    const [pluginUrl, category] = findPlugin(name);
+
+
+    if (pluginUrl === null) {
+        return res.status(404).send('Plugin not found');
```
JMante1, line 17: be clearer: the url called does not correspond to a plugin

lib/plugins/pluginEndpoints
```diff
@@ -0,0 +1,117 @@
...
+        
+    
+        return axios({
+            method: 'GET',
+            url: pluginUrl + 'status',
+            responseType: 'text'
+        }).then(response => {
+            return res.status(200).send(response.data);
+        }).catch(error => {
+            return res.status(400).send('The Plugin is not Responding');
```
JMante1, line 77: The plugin [plugin name] status endpoint is not responding

JMante1 (changes requested):
It looks good. Just could you make sure the error messages are a little more informative. Adding more information as to what error and what plugin etc was called (just in case people think they are calling something else). So in general say what was sent and give a list of options that would work. Or a place to find those options (e.g. run get request to get a list of available plutins)

cjmyers (approved):

cjmyers:
I approve the changes but the docker build failed.  I'm not sure why.  I've restarted it.  We will not be able to merge until the build and tests all complete successfully.

JMante1 (approved):

PR #1525, opened 2023-02-02
Zane dev plugins

zane-perry:


zane-perry:
Note that submit plugins are not entirely functional yet, and I ran across some minor issues with rendering plugins, but error checks are in place that prevent any code from breaking, it just displays an error message in the relevant component

danielfang97 (approved):
Great WORK! :D

PR #1538, opened 2024-04-19
added frontendurl and backendurl to setup, added uriprefix to theme

danielfang97:


cjmyers:
You need to re-run and reset the tests as described here:

See: Changing tests to reflect changes to SynBioHub

This then becomes part of your commit, which I will want to double check

cjmyers (approved):

PR #1563, opened 2024-09-03
Fix dockerignore

PeterHindes:
makes this work as a submodule for dockercompose


PeterHindes:
This is not complete yet. For some reason the docker image needs its git folder >:(

image

PR #1577, opened 2024-11-17
Added config option on theme for require login

cjmyers:


danielfang97 (approved):
Tested and is working.

danielfang97:
Tested with makePublic and once Java is compiled, it's working as intended.

PR #1579, opened 2024-11-22
Fix curator share

cjmyers:
In order to test this one, you need to create two users, one admin/curator and another basic.  The basic one creates a collection with some SBOL objects.  Create a share link.  Then login as the curator and access the collection using the share link.  Then make the collection public.


danielfang97:
Tested and it works. The collection can befound in design collections by both users after publication.

danielfang97 (approved):
Worked. Looks good

PR #1580, opened 2024-11-22
Setuphttps

cjmyers:
To test this, you need to start from scratch and check the URLs guessed on setup.  To really test it, we need to test on a server with a secure URL.


danielfang97 (approved):
Came back with the expected: http://localhost:7777/ on local setup

PR #1582, opened 2024-11-29
Add share sparql endpoint and shareLink endpoint

### LINKED ISSUE #1559: Sparql endpoint needs to properly check access rights to support share links
The sparql endpoint should validate that a user has access to the graphs involved in the query.  For normal requests, this means comparing the URIs to ensure they are public or in users graph.  For share links, need to verify that they are in the URI prefix from the share link, and the hashcode is correct.

cjmyers:
Closes #1559


cjmyers:
Both the GUI and API need to be tested.  To test the GUI, do the following:
1) Create two users (user 1 and user 2)
2) Create a collection for user 1 and get a share link
3) Login as user 2, and go to the share link + /sparql
4) Enter a SPARQL query that uses the collections URI (not share link) as the subject, should get results
5) Go to ordinary SPARQL search page, same query should be blocked from here

To test the API,
1) Login as user 1, get share link using <URI>/shareLink endpoint
2) Login as user 2 to get the token
3) Perform GET request using the SPARQL share link + ?query=<URL encoded query>, should get results
4) Perform GET request using ordinary SPARQL api,  should not get results

danielfang97:
Tested test1 using this query:

SELECT ?predicate ?object
WHERE {
    <http://localhost:7777/user/test1/test1/test1_collection/1> ?predicate ?object .
}

Got the expected results

danielfang97:
image

Did this on postman for testing api but got 404 not found

danielfang97:
Never mind, I was using the wrong login information. It worked this time. Both tests passed, for GUI and API

danielfang97 (approved):
Both tests worked!

cjmyers:
Can you provide more detail of all the API calls you made?

PR #1583, opened 2024-11-29
Check if https used for setup

cjmyers:
Test that setup guess still okay for local host.  Will need to merge to test on server using https


danielfang97 (approved):
tested and remains the same. localhost:7777 on both

PR #1584, opened 2024-11-30
Fix blastp

### LINKED ISSUE #1571: Incorrect BLAST Interface for Protein Sequences
When selecting the BLAST button on a protein sequence, the page incorrectly sends the sequence to perform a BLASTn search instead of a BLASTp search. This issue needs to be addressed so that the sequence is sent to the BLASTp interface as intended.

cjmyers:
Closes #1571


danielfang97 (approved):
Tested and it originally linked to blastn but now links to blastp

PR #1587, opened 2024-12-10
Fix secure setup check

cjmyers:


danielfang97 (approved):
Tested on local and appears to be working

PR #1590, opened 2024-12-14
Change share URI to be derived only from collection

### LINKED ISSUE #1586: Add hash version of /shareLink
This endpoint should use the original share link as the access token, and then send either in the query for a GET request or param for a POST the URI that you want a share link for.

If the object is private and the access token allows access (i.e., in same collection), then it should generate a share link for the URI sent and return that.

Otherwise, it should return an error.

cjmyers:
closes #1586


cjmyers:
Check with me before testing.  This may be a tricky one to test.  It should enable you to complete share link viewing, but I want to be sure it does not break anything else.

cjmyers:
@danielfang97 I've updated this to build the new collectionUri based share link only for the /shareLink endpoint.  This should allow this functionality to support SBH2 share links without changing the behavior of SBH1.  Please test this endpoint, and you can approve it if is working for you.

danielfang97 (commented):
Tested and working. Won't approve yet, since I want to make sure I am understanding the case for this correctly.

danielfang97 (approved):
Working as intended after discussion

PR #1591, opened 2024-12-14
Fix problem with editing types

### LINKED ISSUE #1566: Unable to add circular type
When attempting to add circular type, it does not succeed in adding it.

cjmyers:
closes #1566


danielfang97:
I tried to input in a circular type to a part, but I think I did it wrong, and it threw an error. It can no longer access the part at all. I think when an improper type is added, it should throw an error and try to allow you to change it. Not break the part completely.

danielfang97:
What I'm saying, it should be easier to add/edit a type. People may not know the exact way to do it.

cjmyers:
The inability to access the part after enter a bad type is exactly the error I was trying to fix here.  Could you verify that you tested this on this branch.  I thought I had fixed that behavior.

danielfang97:
Just tested and it appears to no longer throw the error. However, there is a new bug where if you leave the box blank, it remains there. Blank should be treated the same as a wrong type.

cjmyers:
I think that bug was there before too

danielfang97:
So, should I just approve this one? Or are we adding an issue to this PR?

danielfang97 (approved):
Works as expected. Now, it doesn't throw an error

PR #1592, opened 2024-12-14
Restrict login endpoint when public login disallowed

### LINKED ISSUE #1557: Should not be able to access /register endpoint when public account creation disabled

cjmyers:
closes #1557


danielfang97 (approved):
Tested first without changes and /register still accessible. Now, with changes, can't access /register anymore.

PR #1593, opened 2024-12-14
Remove ability to download image

### LINKED ISSUE #1470: Download image does not work
It does not work for VisBOL2, should remove the option from SBH and restore on SBH2.

cjmyers:
closes #1470


danielfang97 (approved):
Download image is no longer available per testing

PR #1594, opened 2024-12-14
Remove create collections from browse page

### LINKED ISSUE #1369: Remove CreateCollection from Browse page
This was removed from other search pages, but is still on the browse page.

cjmyers:
closes #1369


danielfang97 (approved):
Tested, and createCollection option no longer available on browse page

PR #1595, opened 2024-12-14
Change to use mkdirp rather than mkdirp-promise

cjmyers:
Fixes issues with adding attachments


danielfang97:
When is this bug supposed to happen? I tested it, and it appears to be working.

cjmyers:
This was happening when I tried to add any attachments from source.  I think it may be related to the attachment bugs you were experiencing in SBH2.  Basically, mkdirp-promise is broken so changed to mkdirp package.

danielfang97:
What do you mean from source? Is this just doing add attachment on each part page?

cjmyers:
I recall that you were having problems with adding attachments when running from source though working in Docker.  I believe this will fix that issue.  We were using a package that is no longer supported and broken.  To test, you can try to add an attachment using the master branch.  It likely will fail.  However, when run with this branch, it should work.

danielfang97 (approved):
That test worked. It failed on master, but worked on this branch. Approved

PR #1598, opened 2024-12-16
added variables so checkboxes on theme for sbh2 are updated correctly…

danielfang97:
… according to config


cjmyers:
Seems to be failing testing.  Also, not sure about the logic.  Can you explain it?

danielfang97:
Re ran tests, and now it seems to be passing tests.
I did this because for SBH2, it wasn't passing in the variables to SBH1. So lines 50-52 are adding the variables to theme, so they can be accessed by SBH2. 
For the other changes, I am setting the variable based on whether it's true or false. The issue is that I realized that SBH1 does this differently. I think this would be good to discuss tomorrow.

cjmyers:
Rethink and retest the logic for setting the variables.  Remove tests that are not necessary for getting this to pass.

danielfang97:
Closed, as this PR was replaced with PR #1601

PR #1602, opened 2024-12-20
Updated sequence ontology file

### LINKED ISSUE #1416: SO terms
Using the SO term: SO:0002251 causes an error in Synbiohub. The sequence ontology version used might need to be updated

cjmyers:
Closes #1416


danielfang97 (approved):
Tested on both SBH1 and SBH2 and no issues related to sequence ontology now

PR #1604, opened 2024-12-23
Fix issue with remove owner having the wrong link

### LINKED ISSUE #1357: Add owner doesn't check if the owner exists

### LINKED ISSUE #1358: Remove owner is not working

cjmyers:
Closes #1358 
Closes #1357


cjmyers:
Make sure to test adding and removing owners both through the GUI as well as through the API.  The API is critical as once it is working we can add this to SBH2.

danielfang97:
On the GUI, I am not sure how to remove an owner. I can't seem to find the button. I did delete the ownedBy in the otherProperties, but when going to the second user, they were still able to access the part I shared.

danielfang97:
Screenshot 2025-01-08 at 7 26 42 PM

danielfang97:
I did it through the API, and got a success, but when I went to the second user (test1) on the GUI, it still had access as well. I might be doing something wrong, but removeOwner doesn't appear to be working properly.

cjmyers:
Think I found the problem.  Please re-test.

danielfang97:
Tested. Mostly working but still has a couple of bugs.
First, using the GUI, clicked on add owner, and it worked. Clicking on add owner again allows you to add owner again? Not sure if intended. 
Owner 2 does now correctly have access to shared collection. 
Owner 1 removing owner 2 does work, with owner 2 not showing access to shared collection. However, the link that owner 2 used to access the shared collection is still valid.
Another bug is removing owner using the API did not work. Owner 2 still is showing access to the shared collection afterwards.

cjmyers:
Adding twice causes no issues.  It would be difficult to remove from list, so I will leave this.

The share link as we discussed does still work, but shared with me should not show it.

Remove owner from API works, but the userUri needs to be the full user uri and not an id.  This is indeed confusing in the API docs.

PR #1607, opened 2024-12-24
Fixes a number of problems with attachments

### LINKED ISSUE #1389: Option to  specify id for attachment
Give the user an option to specify the ID for an attachment

cjmyers:
Closes #1389 

It add a new field "name" to the /attach endpoint (note: API docs need updating).  This name will be the displayId for the attachment.  This can be specified from the GUI now.

This PR also fixes a number of problems with attachments.  Should test adding file attachments and URL attachments both from the GUI and API.  Should also test attachments added to collections during submission of Zip files.


danielfang97:
Some more bugs. I changed the name. Bug 1 is that the name doesn't update until refreshed. Bug 2 is that once refreshed, the type disappears. Bug 3 is that when downloaded, the downloaded file retains the old name.

danielfang97:
Tested. And no issues with attaching files through API vs GUI. Bugs still remain as before.

danielfang97:
Submitting an attachment as part of a zip file, the name is the long string again

cjmyers:
I think that I've fixed the issues you found.  Please test again.

danielfang97 (approved):
Tested again, and it's now working.

PR #1608, opened 2024-12-24
Prevent circular collections

### LINKED ISSUE #1390: Circuluar dependencies
Direct circular dependencies aren't checked when using addToCollection

cjmyers:
Closes #1390


cjmyers:
Be sure to test both GUI and API

danielfang97 (approved):
Correctly doesn't show up on add to collection list. 
For API, throws a 500 error when attempting to make a circular collection. If that's correct, then should be good to merge.

PR #1616, opened 2025-01-30
Allow open access to users list

cjmyers:
To test this, you should be able to get the user list without being logged in as an admin.  This should allow you to build the add owners list


danielfang97 (approved):
Can access users without admin privileges

Issue #1662, opened 2025-06-18
The docker container needs git


PeterHindes:
An error can be darker compose logs, showing that the synbiohub container is trying to call git to add a tag to the page. We can see that the tag is missing on synbiohub.org

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/b1c34d0f-a8ec-48e5-ab86-558807a34a60" />

PeterHindes:
This might be related:
```/synbiohub # apk add nss git openjdk8-jre openjdk8 maven python alpine-sdk libxml2-dev yarn g++ gcc bash raptor2 jq
ERROR: unable to select packages:
  python (no such package):
    required by: world[python]```

we could be dealing with another repository change issue similar to the one with synbiohub3 frontend

PeterHindes:
see the solution in [this pr](https://github.com/SynBioHub/synbiohub/pull/1665) we get the git commit during docker image build instead of every time the docker container is started

PR #1665, opened 2025-07-02
fix some issues preventing the docker container from starting

PeterHindes:


PeterHindes:
fixes [this issue](https://github.com/SynBioHub/synbiohub/issues/1662)

Gonza10V (changes requested):
The docker image is now building but there is a check that is running for over 3 hours and one failed. Is this expected?

cjmyers:
This is definitely a problem

PeterHindes:
the test failed because it expects the container to have its own health check instead of having it in docker compose, I will add it back and it should pass

PeterHindes:
@cl117 This is now ready for merging. Please review and approve if it looks good.

cjmyers:
Docker is failing to build

docker/healthcheck.js
```diff
@@ -1,10 +0,0 @@
-const axios = require('axios')
-const config = require('./lib/config')
-
-const url = 'http://localhost:' + config.get('port') + '/'
-
-axios.get(url).then(response => {
```
PeterHindes, line 6: I would have to check the docs, but this might not fail on 404 etc

PeterHindes: never mind it does, this should work just fine

PR #1673, opened 2025-07-18
Fix bug caused by checking for user when no one is logged in while vi…

cjmyers:
…ewing public collections


PeterHindes (approved):
Looks good, just use let instead of var like copilot suggested

PR #1740, opened 2026-06-30
Evaluate the use of sbol-db in SynBioHub as a replacement for Virtuoso

marpaia:
This branch makes SynBioHub run on [sbol-db](https://github.com/marpaia/sbol-db) as a drop-in replacement for Virtuoso.

Throughout this process, I realized that Virtuoso has some bugs where it accepts SPARQL which doesn't conform to the SPARQL specification. Rather than reimplementing Virtuoso's bugs in sbol-db, the SynBioHub queries that only worked because Virtuoso was permissive are corrected to valid SPARQL and valid data. They now run correctly on Virtuoso and sbol-db alike.

The branch also adds a triplestore testing harness and a four-way performance benchmark. I think it would be reasonable for this test harness to live in the sbol-db repo if y'all would prefer.

The changes in this branch fall into a few broad categories:

### 1. SynBioHub correctness fixes (valid SPARQL and valid data)

These are real SynBioHub bugs that Virtuoso hid by accepting invalid SPARQL or silently dropping malformed data. A strict, standards-compliant store rejects them. Each fix is valid SPARQL, so it works on both backends.

- `sparql/searchCount.sparql`: move the `FROM` dataset clause out of the inner subquery to the outer query. A dataset clause on a subquery is invalid SPARQL, and on a strict store search returned zero results.
- `lib/fetch/local/fetch-sbol-object-recursive.js`: omit the user-graph `FROM` clause when resolving a purely public object. `graphUri` is null in that case, and emitting `FROM <null>` produces an invalid relative IRI that a strict parser rejects.
- `sparql/AttachUpload.sparql` and `sparql/AttachUrl.sparql`: change `INSERT` to `INSERT DATA` for the ground-triple attachment inserts. A bare `INSERT` with no `DATA` or `WHERE` is invalid, so attaching a file failed and stored nothing.
- `lib/actions/makePublic.js`: credit the publishing user (`req.user.name`) instead of an empty string. Virtuoso silently dropped the empty `dc:creator` literal, while a strict store keeps it, so a blank value appeared in the UI.

### 2. Deterministic query ordering

- `lib/views/admin/graphs.js`: add `ORDER BY ?graph` to the graph-listing query. Without an explicit order the result depended on the store's internal enumeration, which made the admin graphs page non-deterministic across backends and across runs.

### 3. SynBioHub-on-sbol-db test harness

- `tests/sboldb/docker-compose.yml` plus `tests/sboldb/test-sboldb.sh`: one Docker stack runs SynBioHub against the chosen triplestore, then runs the same Python suite the Virtuoso path uses. sbol-db answers at the same hostname and port Virtuoso used, so the SynBioHub config is byte-identical.
- `tests/sboldb/config.local.json`: triplestore endpoints come from config, so one SynBioHub image runs on either backend.
- `--persist` phase: restart the stack with volumes intact and run `test_docker_persist.py` to confirm data survives a container restart.
- `tests/sboldb/run-sboltestrunner.sh`: Java SBOLTestRunner round-trip conformance runner, which also produces the SBOL corpus the benchmark consumes.
- `tests/test_functions.py`: make the server-error-log helper backend-agnostic (via `SBH_TEST_CONTAINER`) and non-fatal, so log retrieval never masks the real test diff.
- `tests/sboldb/README.md`: documents the harness.

### 4. Performance benchmark (four-way)

- `tests/sboldb/bench/`: loads the same SBOL corpus into Virtuoso and all three sbol-db backends (Postgres, SQLite, RocksDB) and times SynBioHub's realized read queries.
- A settle step runs each backend's maintenance pass after ingest and before the read measurements, so reads reflect steady state rather than a freshly bulk-loaded index.
- `gen_report.py` renders LaTeX chart and table fragments for the status-update deck.
- Result: every sbol-db backend beats Virtuoso on all six benchmarked read workloads, with row-count parity against Virtuoso on every workload.

### 5. Trustworthy Virtuoso fixture baseline

The committed fixtures predated the current SynBioHub build, so the suite did not pass on Virtuoso out of the box. Regenerating them against the current Virtuoso image makes every later difference attributable to sbol-db.

- Re-baselined `admin-graphs` (drop Virtuoso system graphs, deterministic sorted order), `browse`, `admin-registries`, and the attachment download XML fixtures, and reverted an `advancedSearch` re-baseline where the duplicate was real.

### 6. Docker build

- `docker/Dockerfile`: install Node dependencies from `package.json` and `yarn.lock` in their own cached layer so application source edits do not re-run `yarn install`. `node_modules` is excluded from the build context.

### Validation (against published `ghcr.io/marpaia/sbol-db:v0.1.1`)

- Python functional suite: all 66 tests pass.
- Persistence check: passes (data survives a container restart).
- Java SBOLTestRunner: 189 of 189 files round-trip.
- Benchmark: row-count parity with Virtuoso on all workloads, and every sbol-db backend faster than Virtuoso on all six read workloads.


marpaia:
Closing in favor of the other PRs we made for this.

